PR #11, opened 2015-05-21
Adding a memcache sliver tool fetcher

mtlynch:
Adding a class that retrieves sliver tools from memcache based on
specified criteria.

### DIFF
--- a/server/mlabns/tests/test_tool_fetcher.py
+++ b/server/mlabns/tests/test_tool_fetcher.py
@@ -30,7 +30,7 @@ class ToolFetcherTestCase(unittest.TestCase):
 
     def testFetchDoesNotHitDatastoreIfMemcacheHasRequiredData(self):
         # The mock response is just ints here for simplicity, though the real
-        # function returns
+        # function returns SliverTool objects.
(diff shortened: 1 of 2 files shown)


server/mlabns/tests/test_tool_fetcher.py
```diff
@@ -40,45 +68,51 @@ def insertSliverTool(self, tool_id, site_id=None, status_ipv4=None,
         tool.latitude = latitude
         tool.longitude = longitude
         tool.country = country
-        tool.put()
+        self.created_tools.append(tool)
```
yachang: So each combination of site and country and IPv4/v6 will have its own Sliver Tool? Where is the code for tools initialization in the real case?

mtlynch: IPv4 and IPv6 are both in the same SliverTool entry, but each slice at each site has its own SliverTool. You can see the full list here:

http://mlab-nstesting.appspot.com/admin/sliver_tools

The information is populated from Nagios with the code in server/mlabns/handlers/update.py

yachang:
LGTM

PR #13, opened 2015-05-26
Refactors resolvers to use the ToolFetcher class

mtlynch:
This simplifies the implementation of the Resolver* classes by moving
the details of memcache/datastore into the ToolFetcher class. This also
includes a mostly complete rewrite of the unit tests for the Resolver
classes, as the previous implementation was very flaky and did not
effectively test the public APIs of the resolver classes.


server/mlabns/tests/test_resolver.py
```diff
@@ -1,490 +1,608 @@
...
+                           _createSliverTool(_TOOL_ID))
+
+        mock_fetch = tool_fetcher.ToolFetcher().fetch
+        mock_fetch.return_value = candidate_tools
+        query_results = self.resolver.answer_query(query)
+
+        # Result should be None when there are no matches.
+        self.assertIsNone(query_results)
+
+    def testAnswerQueryChoosesRandomlyAmoungToolsInCountry(self):
```
yachang: Typo, Among

mtlynch: Good catch! Fixed: 2dfd5f

server/mlabns/tests/test_resolver.py
```diff
@@ -1,490 +1,608 @@
...
+        """When a single tool is closest, return that tool."""
+        query = lookup_query.LookupQuery()
+        query.tool_id = _TOOL_ID
+        query.latitude = 0.0
+        query.longitude = 0.0
+
+        close_tool = _createSliverTool(
+            _TOOL_ID, site_id='abc01', latitude=1.0, longitude=1.0)
+        far_tool = _createSliverTool(
+            _TOOL_ID, site_id='cba01', latitude=5.0, longitude=5.0)
```
yachang: How negative longitudes are handled? for example 175 and -175 are actually quite close.

server/mlabns/tests/test_resolver.py
```diff
@@ -1,490 +1,608 @@
...
+        query.tool_id = tool_id
+
+        # Simulate no matching tools
+        tool_fetcher.ToolFetcher().fetch.return_value = []
+
+        # Result should be None when there are no matches.
+        self.assertIsNone(self.resolver.answer_query(query))
+
+    def testAnswerQueryReturnsRandomToolWhenQueryIsMissingLatLon(self):
+        # TODO(mtlynch): This behavior is confusing because it is inconsistent
```
yachang: missing both or just one?

server/mlabns/tests/test_resolver.py
mtlynch: I think that's out of scope for this pull request as that's the job of server/mlabns/util/distance.py and I'm not touching that code. But it is a good test case and something that's currently not exercised in test_distance.py. Can you file a bug / submit a separate pull request?

server/mlabns/tests/test_resolver.py
mtlynch: This is a test of the scenario when both lat and lon are missing.

yachang:
LGTM

server/mlabns/tests/test_resolver.py
mtlynch: Filed #18

mtlynch:
Thanks for the review!

Issue #9, opened 2015-05-18
Broken links on map page


mtlynch:
(Migrated from code.google.com)
Reported by mtlynch@google.com, Oct 8, 2014
What steps will reproduce the problem?
1. View mlab-ns map view on https://mlab-ns.appspot.com/admin/map/ipv4/ndt

What is the expected output? What do you see instead?
Links at the bottom for "Usage" and "FAQ" are broken. Point to missing pages at:
Usage -> http://www.measurementlab.net/usage
FAQ -> http://www.measurementlab.net/content/faq

I think FAQ is meant to go to: http://www.measurementlab.net/faq

What version of the product are you using? On what operating system?
N/A

Please provide any additional information below.

mtlynch:
Fixed

PR #17, opened 2015-06-01
Fixing broken links

mtlynch:
This fixes the broken links on the dashboard pages of mlab-ns and
also replaces some goo.gl links with their full URLs.


mtlynch:
This fixes issue #9

PR #19, opened 2015-06-03
Updates resolver to use tool_address_family

mtlynch:
Changes the resolver to use the correct property of LookupQuery,
tool_address_family. Also expands unit tests to ensure that this
property is exercised and honored in lookups.


pboothe:
LGTM

mtlynch:
Thanks for the review!

PR #20, opened 2015-06-03
Updates lookup to use tool_address_family

mtlynch:
Updates lookup to use the tool_address_family property of LookupQuery.
Also does some refactoring to eliminate redundant URL/FQDN generation
code and deletes / re-sorts library imports.

### DIFF
--- a/server/mlabns/handlers/lookup.py
+++ b/server/mlabns/handlers/lookup.py
@@ -1,20 +1,68 @@
-from google.appengine.api import memcache
-from google.appengine.api import taskqueue
-from google.appengine.ext import db
-from google.appengine.ext import deferred
-from google.appengine.ext import webapp
-from google.appengine.ext.webapp import template
-
-from mlabns.db import model
-from mlabns.util import constants
-from mlabns.util import message
-from mlabns.util import lookup_query
-from mlabns.util import resolver
-from mlabns.util import util
-
 import json
 import logging
 import time
+
+from mlabns.db import model
+from mlabns.util import lookup_query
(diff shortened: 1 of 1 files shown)


server/mlabns/handlers/lookup.py
```diff
@@ -1,20 +1,68 @@
...
+        address_family: The address family for which to create the URL or None
+            to create an address family agnostic URL.
+        http_port: The HTTP port number part of the URL.
+
+    Returns:
+        A tool URL, annotated with the appropriate address family (when
+        available).
+    """
+    url = 'http://' + _create_af_aware_fqdn(fqdn, address_family)
+    if http_port:
```
pboothe: Is this a check for null or is it genuinely okay to pass a falsy value for `http_port`?

If the latter, then I think we should make the function be:
`def _create_tool_url(fqdn, address_family, http_port=None):`
which makes `http_port` an optional argument.

pboothe:
Except for the one nitpick, LGTM

server/mlabns/handlers/lookup.py
mtlynch: Good point. It's actually a check for null. This made me realize that it's actually illegal to call this function if `http_port` is not set because we shouldn't be constructing HTTP URLs for slices that have no HTTP port assigned (i.e. can't speak HTTP), so I got rid of the `if` in `_create_tool_url` entirely.

Fixed in b20a39ee1f4c57b5091ec350767d1430524e7d7a

mtlynch:
Thanks for the review!

PR #22, opened 2015-06-04
Setting executable bit on environment_bootstrap script

mtlynch:


mtlynch:
Merging without code review as this is not a code change

PR #28, opened 2015-06-11
Fixing None checks for http_port and server_port (again)

mtlynch:
This is a partial revert of b20a39. After testing, I realize we *do*
want implicit false for http_port and server_port, as these are strings,
not ints, as I had incorrectly assumed.


pboothe:
LGTM

Less `is` is more.

mtlynch:
Thanks for the review!

PR #29, opened 2015-06-11
Fixes a regression in geo_options where we only returned 1 tool

mtlynch:
Fixes a regression in geo_options where if the lat/lon was missing from
the LookupQuery, the resolver would return a single random tool instead
of a random subset of tools (geo_options requests should yield a list).

### DIFF
--- a/server/mlabns/util/resolver.py
+++ b/server/mlabns/util/resolver.py
@@ -57,9 +57,6 @@ class GeoResolver(ResolverBase):
     def _get_closest_n_candidates(self, query, max_results):
         """Selects the top N geographically closest SliverTools to the client.
 
-        Finds the top N closest SliverTools to the client and returns them.
-        Note that N is currently hardcoded to 4.
-
         Args:
             query: A LookupQuery instance.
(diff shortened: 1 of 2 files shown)


server/mlabns/util/resolver.py
```diff
@@ -74,8 +74,12 @@ def _get_closest_n_candidates(self, query, max_results):
 
         if (query.latitude is None) or (query.longitude is None):
             logging.warning(
-                'No latitude/longitude, return a random sliver tool.')
-            return [random.choice(candidates)]
+                'No latitude/longitude, return random sliver tool(s).')
+            if len(candidates) <= max_results:
```
pboothe: how about:
`return random.sample(candidates, min(len(candidates), max_results))`

pboothe:
Can be simplified, but the tests look good.  LGTM otherwise.

server/mlabns/util/resolver.py
mtlynch: Good suggestion, fixed!

mtlynch:
Thanks for the review!

PR #31, opened 2015-06-16
Fixes lookup_query to explicitly set default value to None

mtlynch:
Fixes a regression in lookup_query where request.get calls were not
getting explicitly set to default_value=None.

Updates unit tests so that unit test mocks accurately mimic GAE's
request.get:
https://cloud.google.com/appengine/docs/python/tools/webapp/requestclass#Request_get


pboothe:
LGTM

mtlynch:
Thanks!

PR #33, opened 2015-06-16
Making user_defined_country private

mtlynch:
There are no longer external dependencies on user_defined_country so we
can make it private. Additionally, this allows us to treat '' and None
as equivalent for this field.


yachang:
LGTM

mtlynch:
Thanks for the review!

Issue #36, opened 2015-07-09
Do not serve mlab4 slices


mtlynch:
http://mlab-ns.appspot.com/admin/sliver_tools

Includes mlab4 slices, which are not production slices. These should not appear in mlab-ns results and clients should not be routed to these slices.

mtlynch:
Fix is ready, will be deployed to prod next week

PR #37, opened 2015-07-09
Filtering out non-production sites/slices

mtlynch:
This change adds more robust filtering to exclude non-production M-Lab
sites and slices from mlab-ns.

This fixes issue #36.

### DIFF
--- a/server/mlabns/tests/test_production_check.py
+++ b/server/mlabns/tests/test_production_check.py
@@ -1,0 +1,45 @@
+import unittest
+
+from mlabns.util import production_check as pc
+
+
+class CheckProductionTestCase(unittest.TestCase):
+
+    def testIsProductionSite(self):
+        self.assertTrue(pc.is_production_site('nuq01'))
+        self.assertTrue(pc.is_production_site('nuq99'))
+        self.assertTrue(pc.is_production_site('tun05'))
+        self.assertTrue(
+            pc.is_production_site('TUN05'),
+            ('production style site names are considered production even when '
+             'uppercase'))
+
+        self.assertFalse(pc.is_production_site(''))
+        self.assertFalse(pc.is_production_site('foo'))
+        self.assertFalse(
+            pc.is_production_site('lga123'),
+            'Sites with too many numbers are not production sites.')
+        self.assertFalse(
+            pc.is_production_site('lga01t'),
+            'Sites with a t suffix are not production sites.')
+
+    def testIsProductionSlice(self):
+        self.assertTrue(pc.is_production_slice(
+            'ndt.iupui.mlab3.mad01.measurement-lab.org'))
+        self.assertTrue(pc.is_production_slice(
+            '1.michigan.mlab1.hnd01.measurement-lab.org'))
+        self.assertTrue(pc.is_production_slice(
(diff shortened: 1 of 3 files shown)


server/mlabns/util/production_check.py
```diff
@@ -0,0 +1,34 @@
...
+
+def is_production_site(site_name):
+    """Determines if the given site name matches the production site schema
+
+    Args:
+        site_name: Name of site to check (e.g. "lga01")
+
+    Returns:
+        True if the site name matches the schema of a production site.
+    """
```
yachang: So abc09 will pass the test as well?

server/mlabns/tests/test_production_check.py
```diff
@@ -0,0 +1,45 @@
+import unittest
+
+from mlabns.util import production_check as pc
+
+
+class CheckProductionTestCase(unittest.TestCase):
+
+    def testIsProductionSite(self):
+        self.assertTrue(pc.is_production_site('nuq01'))
```
yachang: Add test case that the letter is some non-exist letters, like "zzz09"

server/mlabns/util/production_check.py
```diff
@@ -0,0 +1,34 @@
...
+    return re.match('^[a-z]{3}\d{2}$', site_name, re.IGNORECASE) != None
```
nkinkade: Should this be '^[a-zA-Z{3}\d{2}$'?  In the unit test you test for TUN01, noting that all caps are also valid.

server/mlabns/util/production_check.py
nkinkade: This is probably OK.  We currently only have sites up to 05, but that could change and it might be reasonable to assume that the general form of the string is correct than to make assumptions about how many sites we have deployed in a given region.

server/mlabns/util/production_check.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def is_production_slice(slice_fqdn):
+    """Determines if the given FQDN matches the schema for a production slice
+
+    Args:
+        slice_fqdn: Slice FQDN to check (e.g. "lga01")
```
nkinkade: Tiny detail, since it's only a comment but should the example list an a FQDN instead of a site name?

server/mlabns/util/production_check.py
mtlynch: Correct, `abc09` will pass. According to the schema, we can go up to 99.

server/mlabns/tests/test_production_check.py
mtlynch: `zzz09` should pass because it fits the site schema. All mlab-ns knows is what it sees in ks and what it knows about the schema. If ks says there's a site called `zzz09` then mlab-ns should treat that site as a production site because a) ks reported it and b) it fits the production site schema.

Given this, do you feel like it provides test coverage that the other positive test cases don't provide?

yachang:
LGTM

server/mlabns/util/production_check.py
mtlynch: Good catch! Fixed

mtlynch:
Thanks for the review!

PR #40, opened 2015-07-16
Removing message constants that are no longer used

mtlynch:
[Review on Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/40)


fernandalavalle:
LGTM

PR #43, opened 2015-10-05
Added consistency to the host returned.

pboothe:
Will now always return the lexicographically first host that is available at a
given site instead of randomly selecting one.

### DIFF
--- a/server/mlabns/db/tool_fetcher.py
+++ b/server/mlabns/db/tool_fetcher.py
@@ -46,6 +46,26 @@ def _filter_by_country(tools, country):
     """Filters sliver tools based on the tool's country."""
     return filter(lambda t: t.country == country, tools)
 
+def _filter_choose_one_host_per_site(tools):
+    """Filters to make sure only one host is returned per site_id.
+
+    This filter should be run after _filter_by_status if you want to make sure
+    the chosen site is up.
+
+    Args:
+        tools: The list of tools to filter.
+
+    Returns:
+        A list containing a unique tool for each site.
+    """
+    sites = {}
+    for tool in tools:
+        if tool.site_id not in sites:
+            sites[tool.site_id] = tool
+        else:
+            sites[tool.site_id] = min(sites[tool.site_id], tool,
+                                      key=lambda t: t.fqdn)
+    return [ tool for tool in sites.values() ]
 
 def _find_site_ids_for_metro(metro):
     """Determine which site IDs are present in a given metro.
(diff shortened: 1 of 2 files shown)


server/mlabns/db/tool_fetcher.py
```diff
@@ -46,6 +46,26 @@ def _filter_by_country(tools, country):
...
+        A list containing a unique tool for each site.
+    """
+    sites = {}
+    for tool in tools:
+        if tool.site_id not in sites:
+            sites[tool.site_id] = tool
+        else:
+            sites[tool.site_id] = min(sites[tool.site_id], tool,
+                                      key=lambda t: t.fqdn)
+    return [ tool for tool in sites.values() ]
```
mtlynch: Should be no whitespace before/after square brackets:

https://google-styleguide.googlecode.com/svn/trunk/pyguide.html?showone=List_Comprehensions#List_Comprehensions

server/mlabns/tests/test_tool_fetcher.py
```diff
@@ -87,33 +88,51 @@ def initToolIdSiteGroup(self):
...
-            status_ipv6=message.STATUS_OFFLINE)
+            status_ipv6=message.STATUS_OFFLINE,
+            fqdn='mock_tool_c.mlab1.abc03.measurement-lab.org')
         self.insertCreatedTools()
 
+    def testOnlyReturnMlab1(self):
+        self.initToolIdSiteGroup()
+        tool_properties = tool_fetcher.ToolProperties(tool_id='mock_tool_a')
+        for tool in self.fetcher.fetch(tool_properties):
+          self.assertTrue("mlab2" not in tool.fqdn)
```
mtlynch: I'd prefer to assert that it contains the correct fqdn rather than verifying that it doesn't have the wrong ones. If the filter was broken and was rejecting more than it should, this would still pass.

mtlynch: Single quotes for consistency with the rest of the file.

server/mlabns/tests/test_tool_fetcher.py
```diff
@@ -59,7 +59,7 @@ class ToolFetcherCommonTests(object):
 
     def createSliverTool(self, tool_id, site_id=None, status_ipv4=None,
                          status_ipv6=None, latitude=None, longitude=None,
-                         country=None):
+                         country=None, fqdn=""):
```
mtlynch: Single quotes for consistency with the rest of the file.

mtlynch:
LGTM once that test case is flipped from negative to positive.

server/mlabns/db/tool_fetcher.py
pboothe: Done.

server/mlabns/tests/test_tool_fetcher.py
pboothe: Fixed and fixed.

server/mlabns/tests/test_tool_fetcher.py
pboothe: Done.

PR #46, opened 2015-11-25
Support for querying against the Maxmind database using the pygeoip module.

collina:
[Review on Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/46)


mtlynch:
Cool, this looks really good. A few high level notes:

* Can we switch to Reviewable since I expect this to go a few rounds?
* Can we get unit test(s) for `get_ip_geolocation_datafile`? (mocking out `pygeoip.GeoIP`)
* Can we just strip out all of the datastore-based maxmind code? It's pretty messy to have the dead code sitting alongside the live code.

If we knew that this was just temporary and that we'd be switching back to Datastore in Q1, I'd be more open to leaving both in to reduce code churn, but I think there's a strong possibility that the long term "right" fix is going to get neglected for a while so we'll potentially have dead code and irrelevant tests rotting in the repo for years.

Have you measured the difference in perf (checking out the GCP logs) of this implementation vs. Datastore-based? The only thing that makes this implementation potentially not the right long term approach is that I expect that reading from the filesystem is significantly slower than reading from the Datastore. But if it's a perf hit of like <10%, I think we're better off long term with pygeoip because it's so much less code/complexity.

server/mlabns/util/maxmind.py
```diff
@@ -20,34 +26,76 @@ def __init__(self, city=None, country=None, latitude=None, longitude=None):
...
+        A GeoRecord containing the geolocation data if is found in the db,
+        otherwise an empty GeoRecord.
+
+    Raises:
+        socket.error, if 'remote_addr' is not a valid IP address.
+    """
+    geo_record = GeoRecord()
+
+    geo_city_block = pygeoip.GeoIP(city_file,
+        flags=pygeoip.const.MEMORY_CACHE).record_by_addr(remote_addr)
```
mtlynch: Can pygeoip preserve its cache from request to request on AppEngine? We might be better off with `GEOIP_STANDARD`.

mtlynch:
Can we strip out references to the old MaxMind DataStore data from `server/bulkloader.yaml` and `server/mlabns/db/model.py` as well?

---

Reviewed 2 of 5 files at r1, 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 18 unresolved discussions.

---

*[server/mlabns/tests/test_maxmind.py, line 7 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4OUQ3B8jt9bLqijhE6:-K4OUQ3B8jt9bLqijhE7:-2147429694) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/tests/test_maxmind.py#L7)):*
Two blank lines was right:

>Two blank lines between top-level definitions, be they function or class definitions.

https://google.github.io/styleguide/pyguide.html?showone=Blank_Lines#Blank_Lines

---

*[server/mlabns/tests/test_maxmind.py, line 45 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4OXyKMLdfOQ7DN54Xs:-K4OXyKMLdfOQ7DN54Xt:276863609) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/tests/test_maxmind.py#L45)):*
(optional: not your fault, but now would be a good time for us to fix this) The semantics of `GeoRecord` and its related methods are kind of silly. Consider implementing `__eq__` / `__ne__` in `GeoRecord` so that we can get rid of `assertGeoRecordEqual` and just use normal `self.assertEqual`. 

We could also return `None` instead of an empty GeoRecord (allowing us to test with `self.assertIsNone`) but that's a more aggressive change.

---

*[server/mlabns/tests/test_maxmind.py, line 58 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4PC11xp0dECE_lAzlR:-K4PC11xp0dECE_lAzlS:2003004474) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/tests/test_maxmind.py#L58)):*
I think the preceding 3 tests are actually tests on pygeoip rather than mlab-ns. These are unit tests, so we should just be testing that our code works independently of pygeoip. I recommend cutting/rewriting anything that doesn't mock out pygeoip.

It seems like the question we're asking is "does `get_ip_geolocation` handle it properly when `record_by_addr` returns `None` which is exercised in `testGetGeolocationNoRecordForIp`.

---

*[server/mlabns/tests/test_maxmind.py, line 65 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4PDDIrZWpEafQJOPah:-K4PDDIrZWpEafQJOPai:-210056338) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/tests/test_maxmind.py#L65)):*
I don't think this exercises any of our code. It seems like it's verifying that the `mock` library is working as expected.

---

*[server/mlabns/tests/test_maxmind.py, line 70 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4PHEQq1KCVlEbdFNZD:-K4PHEQq1KCVlEbdFNZE:935675563) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/tests/test_maxmind.py#L70)):*
I think it's important that we see that we're calling `pygeoip.GeoIP` with correct params. Suggest something like:

```
@mock.patch('pygeoip.GeoIP')
    def testGetGeolocationNoRecordForIp(self, mock_geoip_record_by_addr):
        mock_geoip.record_by_addr.return_value = None
        # Check that we initialized with the right backing file and flags                
        mock_geoip.assert_called_with(constants.GEOLOCATION_MAXMIND_CITY_FILE, flags=pygeoip.const.STANDARD)
        # Check that we passed along the correct IP address
        mock_geoip.record_by_addr.assert_called_with('1.2.3.4')`
```

Very likely I'm getting the syntax wrong, but let me know if the intent is unclear.

Ditto for other places we call the mock and it's important we call it with the right params. 

---

*[server/mlabns/tests/test_maxmind.py, line 76 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4PHdr2QaO-7cdAY2DM:-K4PHdr2QaO-7cdAY2DN:1466960791) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/tests/test_maxmind.py#L76)):*
Can we pass in a param that we expect `pygeoip` to actually return a `TypeError` for if we weren't mocking it out?

---

*[server/mlabns/tests/test_maxmind.py, line 84 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4PKBMlmW5_YmF-Xwr0:-K4PKBMlmW5_YmF-Xwr1:246166885) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/tests/test_maxmind.py#L84)):*
In general, I prefer to use obviously dummy values rather than the names of the fields (e.g. `FooCity` instead of `city`). It's hard for a reader to immediately distinguish between `city` the actual field name and `city` the dummy value for a city.

---

*[server/mlabns/util/constants.py, line 13 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4OVmaiAquL7d7cYc1Q:-K4OVmaiAquL7d7cYc1R:1371831584) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/util/constants.py#L13)):*
I'd rather not have to change Python code (trivial as it is) when we update the MaxMind file.

Suggestion:

```
GEOLOCATION_MAXMIND_CITY_FILE = 'mlabns/third_party/maxmind/latest'
```

Then `mlabns/third_party/maxmind/latest` is a symlink to `mlabns/third_party/maxmind/GeoLiteCity_201511.dat`. When we update the file, we can just do file operations and don't have to change Python code (though it will still require a commit).

---

*[server/mlabns/util/maxmind.py, line 18 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4PGwTJs1rQJp64yqsy:-K4PGwTJs1rQJp64yqsz:-1673274256) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/util/maxmind.py#L18)):*
(out of scope, not your problem, but driving me crazy): `CVS`->`CSV`

---

*[server/mlabns/util/maxmind.py, line 28 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4PEK8nZqWijc7Q0Lsv:-K4PEK8nZqWijc7Q0Lsw:307390537) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/util/maxmind.py#L28)):*
Suggest `ip_address` since it doesn't necessarily have to be a non-local IP.

---

*[server/mlabns/util/maxmind.py, line 29 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4PDVgeUEnix_SAX_en:-K4PDVgeUEnix_SAX_eo:-558393151) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/util/maxmind.py#L29)):*
I don't think we should parameterize `city_file` until we have an actual need.

---

*[server/mlabns/util/maxmind.py, line 30 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4OahyIpwrjvn4lIVmM:-K4OahyJNvrzQsnKhtrp:281980662) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/util/maxmind.py#L30)):*
Summary line should fit on a single line: 

>A doc string should be organized as a summary line (one physical line) terminated by a period, question mark, or exclamation point, followed by a blank line, followed by the rest of the doc string starting at the same cursor position as the first quote of the first line. There are more formatting guidelines for doc strings below.

https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments

---

*[server/mlabns/util/maxmind.py, line 34 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4Oat3tPn2VgChCvFto:-K4Oat3tPn2VgChCvFtp:1716087141) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/util/maxmind.py#L34)):*
This says v6 is supported but the summary line seems to imply that this is v4 only

---

*[server/mlabns/util/maxmind.py, line 39 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4PE--VCFXggRSbqwRI:-K4PE--W8fekMnUWPSac:1491097683) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/util/maxmind.py#L39)):*
Suggest
`A populated GeoRecord if matching geolocation data is found for the IP address. Otherwise, an empty GeoRecord.`

---

*[server/mlabns/util/maxmind.py, line 46 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4PExzNYlJ71SA49c2J:-K4PExzO4FdYX-hvojnb:1851568598) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/util/maxmind.py#L46)):*
Can we document what causes these errors? `TypeError` seems like something we should be catching earlier (like how did we even get here with an `int`)? `socket.error` seems weird because I wouldn't expect `pygeoip` to be opening sockets.

---

*[server/mlabns/util/maxmind.py, line 47 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4PFFt3ZOL4EZchaJA_:-K4PFFt3ZOL4EZchaJAa:1311533274) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/util/maxmind.py#L47)):*
Can we add a bit more context? Like `logging.warning('record_by_addr(%s) failed with error: %s', remote_addr, e)`

Also, I'd expect this to be at the level of `logging.error`

---

*[server/mlabns/util/maxmind.py, line 57 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4PFziVqOZfWtnOWubP:-K4PFziVqOZfWtnOWubQ:-907751144) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/util/maxmind.py#L57)):*
(optional) I prefer to avoid nesting logic when possible. Suggest:

```
if not geo_city_block:
    logging.error('IP %s not found in the Maxmind database.', remote_addr)
    return GeoRecord()
    
return GeoRecord(city=geo_city_block['city'],
                 country=geo_city_block['country_code'],
                 latitude = geo_city_block['latitude'],
                 longitude = geo_city_block['longitude'])
```

---

*[server/mlabns/util/maxmind.py, line 57 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4PGnP1CpM88glrXhea:-K4PGnP1CpM88glrXheb:367575212) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/util/maxmind.py#L57)):*
If we got to this point, aren't we guaranteed that `remote_addr` is a str?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-:-K4OaHNehImtY6MFk3BF:-68732078)*

collina:
Review status: 2 of 7 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/mlabns/tests/test_maxmind.py, line 70 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4PHEQq1KCVlEbdFNZD:-K4Qf6rCpElIRaIf8BSW:1489712533) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/tests/test_maxmind.py#L70)):*
I ran into a number of issues attempting to mock the full path of pygeoip.GeoIP, and it became a maze of patches for what wasn't clear return on complexity, this is the only question that I have not addressed. 

---

*[server/mlabns/util/maxmind.py, line 18 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4PGwTJs1rQJp64yqsy:-K4QL6l6L6V-lIe4bIF2:884662448) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/util/maxmind.py#L18)):*
Heh fair.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/mlab-ns/46)*

mtlynch:
Reviewed 3 of 5 files at r3.
Review status: 5 of 7 files reviewed at latest revision, 16 unresolved discussions.

---

*[server/mlabns/tests/test_maxmind.py, line 70 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4PHEQq1KCVlEbdFNZD:-K4UTAKGpiJQxBS4gig8:730450223) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/tests/test_maxmind.py#L70)):*
In general I really don't like touching the filesystem in unit tests. It slows down the test loop and it is an implicit shared state between tests (more of a concern when writing than reading, but still matters somewhat for reading).

I wrote a PoC that I feel is pretty clean:
https://github.com/mtlynch/mlab-ns/commit/13e204b6a6da7d3bf2b307b5153c7d15b3d23ead

Would this work for you?

---

*[server/mlabns/tests/test_maxmind.py, line 76 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4UXW7T7RY7mBdTqI3O:-K4UXW7T7RY7mBdTqI3P:1553368834) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/tests/test_maxmind.py#L76)):*
Why are we deleting this one? I liked this one. I'd add one for `socket.error`.

I don't think we need `testGetGeolocationNotValidAddress`, `testGetGeolocationNonExistentAddress`, or `testGetGeolocationNotValidAddress` because we can do better by defining the exact expectations of pygeoip with mocking. (also two of those tests have the same name)

---

*[server/mlabns/tests/test_maxmind.py, line 23 \[r3\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4UYRLrNiPIzw2iC0To:-K4UYRLrNiPIzw2iC0Tp:-907699478) ([raw file](https://github.com/m-lab/mlab-ns/blob/e63bd4fe96cb7e351939b6896113856fa1c4448a/server/mlabns/tests/test_maxmind.py#L23)):*
Suggest
```
maxmind.GeoRecord(
  city='Greenwich, London',
  country='UK',
  ...
```

---

*[server/mlabns/tests/test_maxmind.py, line 37 \[r3\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4UVY5Ko9LauvdC86lS:-K4UVY5Ko9LauvdC86lT:504830665) ([raw file](https://github.com/m-lab/mlab-ns/blob/e63bd4fe96cb7e351939b6896113856fa1c4448a/server/mlabns/tests/test_maxmind.py#L37)):*
I think dominant convention in mlab-ns  is single quotes

---

*[server/mlabns/tests/test_maxmind.py, line 48 \[r3\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4UYf6jV66JmRoIzF-L:-K4UYf6jV66JmRoIzF-M:716788647) ([raw file](https://github.com/m-lab/mlab-ns/blob/e63bd4fe96cb7e351939b6896113856fa1c4448a/server/mlabns/tests/test_maxmind.py#L48)):*
It feels kind of dirty to catch an assert on an error that the test itself (rather than the production code) is throwing. Suggest either deleting this line or doing something like self.assertTrue(geo_record_1 != geo_record_2)

---

*[server/mlabns/tests/test_maxmind.py, line 104 \[r3\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4UZ65quM_IQ68QlSo2:-K4UZ65quM_IQ68QlSo3:-235997785) ([raw file](https://github.com/m-lab/mlab-ns/blob/e63bd4fe96cb7e351939b6896113856fa1c4448a/server/mlabns/tests/test_maxmind.py#L104)):*
Same as earlier, suggest doing all initialization in the `maxmind.GeoRecord` constructor

---

*[server/mlabns/tests/test_maxmind.py, line 156 \[r3\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4UVm10sPMt5M_lqCgJ:-K4UVm10sPMt5M_lqCgK:1073520158) ([raw file](https://github.com/m-lab/mlab-ns/blob/e63bd4fe96cb7e351939b6896113856fa1c4448a/server/mlabns/tests/test_maxmind.py#L156)):*
Way better, thanks!

---

*[server/mlabns/util/constants.py, line 13 \[r3\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4U32oPN-jJ4PJ_-kP0:-K4U32oPN-jJ4PJ_-kP1:-215112747) ([raw file](https://github.com/m-lab/mlab-ns/blob/e63bd4fe96cb7e351939b6896113856fa1c4448a/server/mlabns/util/constants.py#L13)):*
Can we include this file in the commit?

---

*[server/mlabns/util/maxmind.py, line 18 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4PGwTJs1rQJp64yqsy:-K4U37BRCP3SdtPcjeV8:994769105) ([raw file](https://github.com/m-lab/mlab-ns/blob/27a70d711ca67c7c3b06833c35d53e49f08996c3/server/mlabns/util/maxmind.py#L18)):*
Actually I realize this is in scope and needs to be updated because we're not using the CSV files anymore.

---

*[server/mlabns/util/maxmind.py, line 14 \[r3\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4UVgiOQP03Svby2pqS:-K4UVgiOQP03Svby2pqT:1207811914) ([raw file](https://github.com/m-lab/mlab-ns/blob/e63bd4fe96cb7e351939b6896113856fa1c4448a/server/mlabns/util/maxmind.py#L14)):*
Single quotes

---

*[server/mlabns/util/maxmind.py, line 38 \[r3\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4UTni47oSYvTaE3JCk:-K4UTni47oSYvTaE3JCl:146913999) ([raw file](https://github.com/m-lab/mlab-ns/blob/e63bd4fe96cb7e351939b6896113856fa1c4448a/server/mlabns/util/maxmind.py#L38)):*
Suggest `return not self.__eq__(other)`

---

*[server/mlabns/util/maxmind.py, line 52 \[r3\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4UZM4XShFlo8zzaw2i:-K4UZM4XShFlo8zzaw2j:-978812099) ([raw file](https://github.com/m-lab/mlab-ns/blob/e63bd4fe96cb7e351939b6896113856fa1c4448a/server/mlabns/util/maxmind.py#L52)):*
throw -> throws

---

*[server/mlabns/util/maxmind.py, line 58 \[r3\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4UUYxXbWchL8rl-Eer:-K4UUYxXbWchL8rl-Ees:-2094908762) ([raw file](https://github.com/m-lab/mlab-ns/blob/e63bd4fe96cb7e351939b6896113856fa1c4448a/server/mlabns/util/maxmind.py#L58)):*
Can we make this a `#` comment and write as prose? Formatting it with `Raises:` confused me because I thought I was reading a function docstring (esp because this is 1 line away from an actual function docstring)

---

*[server/mlabns/util/maxmind.py, line 61 \[r3\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4UUTTYif7KVV031Vtk:-K4UUTTYif7KVV031Vtl:1975181775) ([raw file](https://github.com/m-lab/mlab-ns/blob/e63bd4fe96cb7e351939b6896113856fa1c4448a/server/mlabns/util/maxmind.py#L61)):*
`exception_error` seems kind of redundand. `error`? `e`?

---

*[server/mlabns/util/maxmind.py, line 62 \[r3\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4UZUUVql3wxkeJMk3q:-K4UZUUVql3wxkeJMk3r:408850282) ([raw file](https://github.com/m-lab/mlab-ns/blob/e63bd4fe96cb7e351939b6896113856fa1c4448a/server/mlabns/util/maxmind.py#L62)):*
single quotes

---

*[server/mlabns/util/maxmind.py, line 73 \[r3\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4UVOEoh2H-62DD_kM5:-K4UVOEoh2H-62DD_kM6:271582995) ([raw file](https://github.com/m-lab/mlab-ns/blob/e63bd4fe96cb7e351939b6896113856fa1c4448a/server/mlabns/util/maxmind.py#L73)):*
Should be no spaces around ` = `

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/mlab-ns/46)*

mtlynch:
Reviewed 2 of 5 files at r3, 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[server/mlabns/util/maxmind.py, line 61 \[r3\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-K4UUTTYif7KVV031Vtk:-K4UngxSimOQpDSS2ovD:-935736440) ([raw file](https://github.com/m-lab/mlab-ns/blob/e63bd4fe96cb7e351939b6896113856fa1c4448a/server/mlabns/util/maxmind.py#L61)):*
It's not really an `exception_message`. When we print it, we see the message, but it's an exception that contains a message property.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/mlab-ns/46)*

mtlynch:
LGTM

---

Review status: all files reviewed at latest revision, 11 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/mlab-ns/46#-:-K4Uo0jRqnhtcg043Pqt:676834233)*

Issue #48, opened 2015-12-01
Remove workaround for ndt-ssl FQDN


mtlynch:
Long term, we plan to migrate all of our slice names to have M-Lab hosts use dashes instead of dots for parts of the hostname before `measurement-lab.org`. That is:

`ndt.iupui.mlab1.lga06.measurement-lab.org` -> `ndt-iupui-mlab1-lga06.measurement-lab.org`

ndt-ssl will be using this naming scheme first, so we temporarily have to special case ndt-ssl.

In the long term we should remove this special-case workaround. This bug tracks the work item to remove the workaround for ndt-ssl.

nkinkade:
This is no longer really a special case. We convert the dotted names to dashed names for `ndt_ssl`, `ndt7` and `neubot_tls`.  It will probably be a very, very long time before we migrate to fully away from dotted names.

PR #51, opened 2015-12-02
Adds support for NDT-SSL.

mtlynch:
* Adds a special workaround for NDT-SSL as this tool is a special case
with respect to FQDNs. If the client queries for NDT-SSL, we first
convert the FQDN's subdomain to use dash separators rather than dot
separators (i.e. there is a single subdomain before the
measurement-lab.org portion of the FQDN). See #48 for more details.
* Adds an ndt_ssl tool to the tools config.

[Review on Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/51)

### DIFF
--- a/server/mlabns/conf/tools.csv
+++ b/server/mlabns/conf/tools.csv
@@ -1,6 +1,6 @@
 tool_id,slice_id,http_port,show_tool_extra
 ndt,iupui_ndt,7123,False
-ndt_ssl,iupui_ndt,7123,False
+ndt_ssl,iupui_ndt,,False
 mobiperf,michigan_1,,False
 npad,iupui_npad,8000,False
 ooni,mlab_ooni,,True
--- a/server/mlabns/conf/tools.csv
+++ b/server/mlabns/conf/tools.csv
@@ -1,5 +1,6 @@
 tool_id,slice_id,http_port,show_tool_extra
 ndt,iupui_ndt,7123,False
+ndt_ssl,iupui_ndt,7123,False
 mobiperf,michigan_1,,False
 npad,iupui_npad,8000,False
 ooni,mlab_ooni,,True
--- a/server/mlabns/handlers/lookup.py
+++ b/server/mlabns/handlers/lookup.py
@@ -3,6 +3,7 @@ import logging
 import time
 
 from mlabns.db import model
+from mlabns.util import fqdn_rewrite
 from mlabns.util import lookup_query
 from mlabns.util import message
 from mlabns.util import resolver
@@ -12,54 +13,17 @@ from google.appengine.ext import webapp
 from google.appengine.ext.webapp import template
 
 
-def _create_af_aware_fqdn(fqdn, address_family):
-    """Adds the v4/v6 only annotation to the fqdn.
-
-    Example:
-        fqdn:       'npad.iupui.mlab3.ath01.measurement-lab.org'
-        ipv4 only:  'npad.iupui.mlab3v4.ath01.measurement-lab.org'
-        ipv6 only:  'npad.iupui.mlab3v6.ath01.measurement-lab.org'
-
-    Args:
-        fqdn: A tool FQDN with no address family specific annotation.
-        address_family: The address family for which to create the FQDN or None
-            to create an address family agnostic FQDN.
-
-    Returns:
-        A FQDN specific to a particular address family, or the original FQDN
-        if an address family is not specified.
-    """
-    if address_family == message.ADDRESS_FAMILY_IPv4:
-        fqdn_annotation = 'v4'
-    elif address_family == message.ADDRESS_FAMILY_IPv6:
-        fqdn_annotation = 'v6'
-    elif not address_family:
-        fqdn_annotation = ''
-    else:
-        logging.error('Unrecognized address family: %s', address_family)
-        return fqdn
-
-    fqdn_parts = fqdn.split('.')
-    fqdn_parts[2] += fqdn_annotation
-
-    return '.'.join(fqdn_parts)
-
-
-def _create_tool_url(fqdn, address_family, http_port):
+def _create_tool_url(fqdn, http_port):
     """Create an HTTP URL for a tool.
 
     Args:
-        fqdn: A tool FQDN with no address family specific annotation.
-        address_family: The address family for which to create the URL or None
-            to create an address family agnostic URL.
+        fqdn: A tool FQDN (already processed by fqdn_rewrite).
         http_port: The HTTP port number part of the URL.
 
     Returns:
-        A tool URL, annotated with the appropriate address family (when
-        available).
+        A constructed tool HTTP URL with protocol and http port.
     """
-    return 'http://%s:%s' % (_create_af_aware_fqdn(fqdn, address_family),
-                             http_port)
+    return 'http://%s:%s' % (fqdn, http_port)
 
 
 class LookupHandler(webapp.RequestHandler):
@@ -123,8 +87,9 @@ class LookupHandler(webapp.RequestHandler):
 
         bt_data = "";
         for sliver_tool in sliver_tools:
-            fqdn = _create_af_aware_fqdn(sliver_tool.fqdn,
-                                         query.tool_address_family)
+            fqdn = fqdn_rewrite.rewrite(sliver_tool.fqdn,
+                                        query.tool_address_family,
+                                        sliver_tool.tool_id)
 
             data = sliver_tool.city
             data += ", "
@@ -175,15 +140,15 @@ class LookupHandler(webapp.RequestHandler):
                 if sliver_tool.status_ipv6 == message.STATUS_ONLINE:
                     ips.append(sliver_tool.sliver_ipv6)
 
+            fqdn = fqdn_rewrite.rewrite(sliver_tool.fqdn,
+                                        query.tool_address_family,
+                                        sliver_tool.tool_id)
             if sliver_tool.http_port:
-                data['url'] = _create_tool_url(sliver_tool.fqdn,
-                                               query.tool_address_family,
-                                               sliver_tool.http_port)
+                data['url'] = _create_tool_url(fqdn, sliver_tool.http_port)
             if sliver_tool.server_port:
                 data['port'] = sliver_tool.server_port
 
-            data['fqdn'] = _create_af_aware_fqdn(sliver_tool.fqdn,
-                                                 query.tool_address_family)
+            data['fqdn'] = fqdn
             data['ip'] = ips
             data['site'] = sliver_tool.site_id
             data['city'] = sliver_tool.city
@@ -242,8 +207,10 @@ class LookupHandler(webapp.RequestHandler):
         sliver_tool = sliver_tools[0]
 
         if sliver_tool.http_port:
-            url = _create_tool_url(sliver_tool.fqdn, query.tool_address_family,
-                                   sliver_tool.http_port)
+            fqdn = fqdn_rewrite.rewrite(sliver_tool.fqdn,
+                                        query.tool_address_family,
+                                        sliver_tool.tool_id)
+            url = _create_tool_url(fqdn, sliver_tool.http_port)
             return self.redirect(url)
 
         return util.send_not_found(self, 'html')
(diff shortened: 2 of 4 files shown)


server/mlabns/util/fqdn_rewrite.py
```diff
@@ -0,0 +1,102 @@
...
+        ndt.iupui.mlab1.lga06.measurement-lab.org
+    NDT-SSL uses:
+        ndt-iupui-mlab1-lga06.measurement-lab.org
+
+    The long term plan is to adjust all FQDNs to match this schema, but
+    currently Nagios (the source of mlab-ns' FQDN data) is not are not
+    aware of the dash-separator schema. We apply a special-case workaround here
+    so that we serve the correct FQDNs to clients for NDT-SSL queries until we
+    fix the naming in Nagios.
+
```
nkinkade: Is the above comment true? I was under the impression that the reasoning for this workaround was so that we could use an SSL wildcard cert against *.measurement-lab.org, not because of some limitation or unimplemented functionality in Nagios.

On additional small note: while we may look to implement this "flat" domain naming for all experiments, I think the plan is to pretty much always leave the old RRs in place for legacy reasons.

server/mlabns/util/fqdn_rewrite.py
```diff
@@ -0,0 +1,102 @@
...
+        fqdn: A tool FQDN with no address family specific annotation.
+        address_family: The address family for which to create the FQDN or None
+            to create an address family agnostic FQDN.
+        tool_id: Name of tool associated with the FQDN (e.g. 'ndt_ssl').
+
+    Returns:
+        FQDN after rewriting to apply all modifications to the raw FQDN.
+    """
+    rewritten_fqdn = _apply_af_awareness(fqdn, address_family)
+    rewritten_fqdn = _apply_ndt_ssl_workaround(rewritten_fqdn, tool_id)
```
nkinkade: Would it be better to wrap this in the conditional if tool_id == "ndt_ssl" rather than passing non NDT-SSL tools to a function named _apply_ndt_ssl_workaround(), and then doing the conditional in the that function?  Adding the conditional here is higher up in the chain and I think more readable, and will not pass control of the code to a function that shouldn't be working on the data its given in any case.

server/mlabns/util/fqdn_rewrite.py
```diff
@@ -0,0 +1,102 @@
...
+    Args:
+        fqdn: A tool FQDN with no address family specific annotation.
+        address_family: The address family for which to create the FQDN or None
+            to create an address family agnostic FQDN.
+        tool_id: Name of tool associated with the FQDN (e.g. 'ndt_ssl').
+
+    Returns:
+        FQDN after rewriting to apply all modifications to the raw FQDN.
+    """
+    rewritten_fqdn = _apply_af_awareness(fqdn, address_family)
```
nkinkade: NOTE: AF domain name decorations are _not_ currently implemented nor published.  We _don't_ want to decorate domain names for NDT-SSL with AF until or if we actually publish such records.   @stephen-soltesz may have more insight into why since he implemented the change to create the "flat" names.  I suspect it's because we don't currently really run NDT over IPv6.  However, I believe we did enable IPv6 for [a select couple nodes](https://github.com/m-lab/operator/commit/523e54a38296d9180d04f1af48a7470b0e1ef48a) 5 or 6 months ago, specifically nuq02 and lga06.

server/mlabns/conf/tools.csv
```diff
@@ -1,5 +1,6 @@
 tool_id,slice_id,http_port,show_tool_extra
 ndt,iupui_ndt,7123,False
+ndt_ssl,iupui_ndt,7123,False
```
nkinkade: fakewww hasn't been updated to support SSL, returning fqdn:7123 won't make sense in the context of ndt_ssl, at least not until fakewww is also updated.  It might be better to populate the http_port field with the expected SSL port of 3010.  Also see note below about creating about the http:// protocol in _create_tool_url()

server/mlabns/handlers/lookup.py
```diff
@@ -12,54 +13,17 @@
...
         http_port: The HTTP port number part of the URL.
 
     Returns:
-        A tool URL, annotated with the appropriate address family (when
-        available).
+        A constructed tool HTTP URL with protocol and http port.
     """
-    return 'http://%s:%s' % (_create_af_aware_fqdn(fqdn, address_family),
-                             http_port)
+    return 'http://%s:%s' % (fqdn, http_port)
```
nkinkade: If this happens to be an NDT-SSL request, then prepending a protocol of http won't be correct.  At the very least it should probably be https.  It could potentially also be wss, which would be useful for Websocket clients.  I don't believe there are any existing full NDT clients which support SSL other than the Websocket one, though I believe Peter threw together a small, rough C one for testing.  If we don't pass a protocol of wss for ndt_ssl requests, then it will be incumbent upon the client to substitute https or http (or whatever) with the appropriate protocol to work around this limitation of mlab-ns.

server/mlabns/util/fqdn_rewrite.py
mtlynch: Yeah, I jumped the gun on this a bit, probably not ready to be codified into the documentation until we flesh out the plan.

Rewritten to be less specific about future plans.

server/mlabns/conf/tools.csv
mtlynch: I don't think fakewww behavior should be different between NDT and NDT-SSL. We might one day offer a fakewww with TLS support for people who want to embed the NDT websockets frontend, but I see that as independent from this change.

server/mlabns/handlers/lookup.py
mtlynch: Similar to the above, the fakewww interface is still the same. We haven't added TLS support to fakewww, so an https URL wouldn't work. This is the fakewww for an NDT-SSL server:

http://ndt.iupui.mlab4.nuq1t.measurement-lab.org:7123/

fakewww only speaks HTTP, not NDT so fakewww's support for TLS is independent of an actual NDT test's support of TLS.

server/mlabns/util/fqdn_rewrite.py
mtlynch: Oh, interesting. I think this is the right change for right now because the AF-annotation is just a refactoring that preserves existing behavior.

Are you sure we don't yet support IPv6? I can query mlab-ns right now and get an v6-annotated FQDN that resolves:

```
$ curl http://mlab-ns.appspot.com/npad?address_family=ipv6
{"city": "New York_NY", "url": "http://npad.iupui.mlab1v6.lga05.measurement-lab.org:8000", "ip": ["2001:1900:2100:14::12"], "fqdn": "npad.iupui.mlab1v6.lga05.measurement-lab.org", "site": "lga05", "country": "US"}

$ curl http://npad.iupui.mlab1v6.lga05.measurement-lab.org:8000
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">

<html>
<head>
  <title>NPAD Diagnostics at Measurement Lab (New York, NY, US)</title>
  <style type="text/css">
...
`

server/mlabns/util/fqdn_rewrite.py
mtlynch: Sure, fixed.

server/mlabns/conf/tools.csv
mtlynch: Oh, I see what you're saying. It doesn't make sense to have NDT-SSL advertise fakewww because it will just direct clients to NDT-plaintext. Alright, I think it makes sense to just have nothing here because if a client wants NDT-SSL and wants to get to it through a web portal, fakewww doesn't help them, so we might as well say there's no web portal for NDT-SSL at the moment. `3010` would be confusing because this is a port that we're telling clients can speak HTTP and NDT-SSL can't speak HTTP.

Fixed, good catch.

server/mlabns/handlers/lookup.py
mtlynch: Okay, after realizing you're right, I wanted to revisit this. It should remain http because all of our tools still only offer web interfaces in http.

server/mlabns/util/fqdn_rewrite.py
nkinkade: Sorry, I may not have explained myself well enough.  Let's say that someone queries mlab-ns using the ndt_ssl tool id:

http://mlab-ns.appspot.com/ndt_ssl?address_family=ipv4

... and is, say, returned this rewritten host name:

ndt-iupui-mlab2v4-mia01.measurement-lab.org

Name resolution will fail.  "flattened", AF decorated DNS RRs do not currently exist.  My comment here was just to make sure that if someone is querying for ndt_ssl that the fqdn is not passed through _apply_af_awareness().

server/mlabns/handlers/lookup.py
nkinkade: I suppose this makes sense, since we don't currently have an SSL-enabled fakewww.  However, how does someone discover the port on which NDT-SSL is listening, or do they just have to know?  The same issue exists for non-SSL queries too, but I suppose we've overlooked it because convention has specified that the port is 3001 since the early days.  Also, it should be noted, for example, that the following URL does give you a secure connection in a browser, but just doesn't run a test properly:

https://ndt-iupui-mlab4-nuq1t.measurement-lab.org:3001/

If we don't care about advertising the SSL port in any way, then I guess the best thing to do is simply leave the non-SSL fakewww URL, despite the fact that it's a little confusing to return non-SSL resources for a specifically SSL request.

mtlynch: > Also, it should be noted, for example, that the following URL does give you a secure connection in a browser, but just doesn't run a test properly:

>https://ndt-iupui-mlab4-nuq1t.measurement-lab.org:3001/

Right this is expected. The browser will connect to 3001 and do a TLS handshake, then try to speak HTTP. The server will accept a TLS handshake, then try to speak NDT, so you get undefined behavior because they're talking to each other over different protocols. I just tried this with Chrome on Ubuntu and the browser thinks the server is serving a small binary file and downloads it.

>If we don't care about advertising the SSL port in any way, then I guess the best thing to do is simply leave the non-SSL fakewww URL, despite the fact that it's a little confusing to return non-SSL resources for a specifically SSL request.

I respectfully disagree. If we have a client that requests a web portal for `/ndt-ssl` it doesn't make sense to direct them to a web portal for ndt plaintext. We don't have a web portal for NDT-SSL, and so I think it's accurate for us to omit the `http_port` in the response for `/ndt-ssl` because omitting that means "we don't have a web interface for this server."

server/mlabns/util/fqdn_rewrite.py
mtlynch: Ah, okay, I see what you're saying.

How much work would it be for us to publish AF-decorated RRs for NDT-SSL?

I can apply another special case in `fqdn_rewrite.rewrite()` but I'd like to limit special casing if possible because it makes it much easier for bugs to crop up when we forget that some other piece of the system assumes no special casing. So if this is something we're going to end up doing in the future, I'd prefer to do it now.

I actually know nothing about why we publish AF-decorated RRs in the first place. I just found that behavior defined in the mlab-ns spec and tried to make sure we preserved it.

server/mlabns/handlers/lookup.py
nkinkade: I don't think we are disagreeing here.  My original point, spread across two comments, was that returning http:7123 to an "ndt_ssl" request didn't seem right.  Removing the port 7123 from the URL is a start, but still returning an http "url" (even minus the port) is also misleading, because the it's not even a valid URL.  That is, the browser will try to connect on port 80, which will fail.  I think we shouldn't publish any "url" at all for "ndt_ssl" requests until such a time as we have an actual SSL url that works.

Are we saying the same thing here, and have just got lost in the chain of comments?

mtlynch: Ah! Cool, yeah we agree. The thing we were missing was that I neglected to mention that mlab-ns only publishes a URL if `http_port` is defined, so it won't report any URL for `/ndt_ssl` responses.

nkinkade:
With m-lab/salt-config-private#2 about to be in place, LGTM!

PR #53, opened 2015-12-04
Fixing IPUpdateHandler to be compatible with slices that share FQDNs

mtlynch:
mlab-ns assumes that FQDNs are unique across slices. This is no longer true as of NDT-SSL, as NDT-SSL slices share FQDNs with NDT slices.

This fixes `IPUpdateHandler` to relax the assumption that FQDNs are unique across slices.

**Note**: This is mostly whitespace changes. There are only 3 lines of actual changes. For best viewing, read in a diff viewer that can ignore non-whitespace changes.

[Review on Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/53)


pboothe:
LGTM.  I was thrown by all the changes, but reviewable.io helpfully highlights the whitespace changes a darker green and the minimality of the change was then clear.

PR #57, opened 2016-02-08
Fetch mlab-host-ips.txt and mlab-site-stats.json from Nagios, not KS

nkinkade:
Our Nagios instance is now running on a Linux VM, not KS. However, two important files were still being fetches from URLs at ks.measurementlab.net.  This PR updates those URLs to use nagios.measurementlab.net, which is more generic and portable in any case.  It also cleans up comments and variable names to use "nagios" instead of "ks".

[Review on Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/57)


mtlynch:
LGTM

PR #61, opened 2016-02-16
Add yapf check.

fernandalavalle:
Added Travis CI with yapf checks. 
Style formatting only applied to /server/mlabns/ excluding /server/mlabns/third_party/. 
Also made all the style changes to make yapf happy.

[Review on Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/61)


.travis.yml
```diff
@@ -0,0 +1,7 @@
+language: python
+python:
+  - "2.7"
+install:
+    - pip install yapf==0.6.2
+script:
+    - yapf --diff --exclude=./server/mlabns/third_party/* --recursive --style=google ./server/mlabns/*
```
mtlynch: Can we add an equivalent pre-commit.sh as well (like in [telescope](https://github.com/m-lab/telescope/blob/development/dev/pre-commit.sh))?

server/mlabns/handlers/lookup.py
```diff
@@ -85,7 +85,7 @@ def send_bt_response(self, sliver_tools, query):
             logging.error("Problem: sliver_tools is not a list.")
             return
 
-        bt_data = "";
```
mtlynch: lol

server/mlabns/tests/test_resolver.py
```diff
@@ -392,24 +428,33 @@ def testAnswerQueryWhenFourToolsAreEquallyClosest(self):
...
+                              latitude=1.0,
+                              longitude=1.0), _createSliverTool(_TOOL_ID,
+                                                                site_id='abc04',
+                                                                latitude=1.0,
+                                                                longitude=1.0),
+            _createSliverTool(_TOOL_ID,
+                              site_id='cba01',
+                              latitude=5.0,
+                              longitude=5.0)
+        ]
```
mtlynch: ditto about suppressing YAPF here

server/mlabns/tests/test_resolver.py
```diff
@@ -423,21 +468,32 @@ def testAnswerQueryWhenMoreThanFourToolsAreEquallyClosest(self):
...
+                              latitude=1.0,
+                              longitude=1.0), _createSliverTool(_TOOL_ID,
+                                                                site_id='abc06',
+                                                                latitude=1.0,
+                                                                longitude=1.0),
+            _createSliverTool(_TOOL_ID,
+                              site_id='cba01',
+                              latitude=5.0,
+                              longitude=5.0)
+        ]
```
mtlynch: basically all of them in this file

.travis.yml
```diff
@@ -0,0 +1,7 @@
+language: python
+python:
+  - "2.7"
+install:
+    - pip install yapf==0.6.2
+script:
```
mtlynch: > I actually don't think we can? I used the slash the way you would in python and the travis validator returned a syntax error. let me know if there's another way of doing it?

Oh, right it's YAML. Does this work?

```yaml
script:
     - |
      yapf --diff --exclude=./server/mlabns/third_party/* --recursive --style=google \
        ./server/mlabns/*
```

fernandalavalle: I think that's only for strings. Validator still claims syntax error with "|" and ">". I looked through the other mlab repos and couldn't find any multiline commands and I couldn't find anything answering about non string multiline commands through just google searching.

server/mlabns/tests/test_resolver.py
fernandalavalle: so just anytime we call _createSliverTool? other formatting seems fine to me

.travis.yml
fernandalavalle: put it in a folder server/mlabns/dev. any important logic to placement?

mtlynch: Suggest just `dev/` or `hooks/` at the root directory. The only goal in placement I think is making it obvious and easy to developers who work on mlab-ns to use the hooks.

server/mlabns/tests/test_resolver.py
mtlynch: I don't think it's this function in particular, but I notice YAPF gets weird when there's Python lists of function calls.

We want to disable YAPF any time it formats the code in a way that looks bizarre. This to me looks pretty weird and not really like what a human would write.

.travis.yml
mtlynch: I think there has to be some way of doing it, but I don't think anyone on the team knows YAML very well. Just a minor thing though, so it's not worth spending much time if the fix isn't easy.

dev/pre-commit.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+yapf --diff --exclude=../server/mlabns/third_party/* --recursive --style=google \
+	../server/mlabns/*
```
mtlynch: Can we replace the tab with spaces?

mtlynch:
Reviewed 26 of 28 files at r1, 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/mlab-ns/61)*

mtlynch:
LGTM

PR #63, opened 2016-02-19
Add unittests to Travis CI

fernandalavalle:
Modified .travis.yml to run unittests check automatically.

[Review on Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/63)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,14 @@
 language: python
 python:
- - "2.7"
-cache:
-	directories:
-	- "$HOME/google-cloud-sdk/"
+  - "2.7"
 env:
-	# Make sure gcloud command is on our PATH and the App Engine SDK is in the Python path
-	- GAE_PYTHONPATH=${HOME}/.cache/google_appengine PATH=$PATH:${HOME}/google-cloud-sdk/bin PYTHONPATH=${PYTHONPATH}:${GAE_PYTHONPATH} CLOUDSDK_CORE_DISABLE_PROMPTS=1 
+  - GAE_PYTHONPATH=$HOME/google_appengine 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,8 +1,15 @@
 language: python
 python:
-  - "2.7"
+ - "2.7"
```
mtlynch: Can we bring back the indenting?

.travis.yml
```diff
@@ -1,8 +1,15 @@
...
+env:
+- PATH=$PATH:$HOME/google_appengine GAE_PYTHONPATH=$HOME/google_appengine
```
mtlynch: Is updating the `PATH` variable necessary?

.travis.yml
```diff
@@ -1,8 +1,15 @@
 language: python
 python:
-  - "2.7"
+ - "2.7"
+env:
+- PATH=$PATH:$HOME/google_appengine GAE_PYTHONPATH=$HOME/google_appengine 
+before_install: 
+- if [ ! -d $HOME/google_appengine ]; then wget https://storage.googleapis.com/appengine-sdks/featured/google_appengine_1.9.32.zip; fi
```
mtlynch: The directory definitely won't be there, so we don't need to do the `if` check.

mtlynch:
There are a lot of commits here (most of which I think are non-functional?). Can we collapse these down to a single commit just to reduce the amount of noise in the commit history (with `git reset --soft`)? 1-2 extra it usually not a big deal, but this one has 14 commits for only 13 lines of changes.

mtlynch:
LGTM

PR #64, opened 2016-02-22
Added pyflakes and fixed what it pointed out

fernandalavalle:
Incorporated pyflakes into Travis CI and corrected all the unused imports and unused assignments it pointed it.

[Review on Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/64)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,14 @@ before_install:
   - wget https://storage.googleapis.com/appengine-sdks/featured/google_appengine_1.9.32.zip
   - unzip -q google_appengine_1.9.32.zip -d $HOME
 install:
+  - pip install pyflakes
   - pip install unittest2 
   - pip install yapf==0.6.2
 script:
   - yapf --diff --exclude=./server/mlabns/third_party/* --recursive --style=google ./server/mlabns/*
   - python server/unit_tests.py $GAE_PYTHONPATH server/mlabns/tests/
+  - PYFILE_LIST=$(find ./server/mlabns/ | grep .py$ | grep -v third_party)
+  - pyflakes $PYFILE_LIST
+
+
+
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -7,8 +7,14 @@ before_install:
   - wget https://storage.googleapis.com/appengine-sdks/featured/google_appengine_1.9.32.zip
   - unzip -q google_appengine_1.9.32.zip -d $HOME
 install:
+  - pip install pyflakes
   - pip install unittest2 
   - pip install yapf==0.6.2
 script:
   - yapf --diff --exclude=./server/mlabns/third_party/* --recursive --style=google ./server/mlabns/*
   - python server/unit_tests.py $GAE_PYTHONPATH server/mlabns/tests/
+  - PYFILE_LIST=$(find ./server/mlabns/ | grep .py$ | grep -v third_party)
```
mtlynch: We need to escape the `.`, so suggest `grep "\.py$"`

Also maybe add a trailing slash to third_party to make it obvious it's only for that directory

.travis.yml
```diff
@@ -7,8 +7,14 @@ before_install:
...
+  - pyflakes $PYFILE_LIST
+
+
+
```
mtlynch: Can we delete the extra newlines?

server/mlabns/handlers/admin.py
```diff
@@ -1,17 +1,13 @@
...
 
 from google.appengine.api import memcache
-from google.appengine.ext import db
 from google.appengine.ext import webapp
 from google.appengine.ext.webapp import template
 
 from mlabns.db import model
 from mlabns.util import constants
-from mlabns.util import message
-from mlabns.util import util
```
mtlynch: Should be no newline separating these

mtlynch:
LGTM

PR #66, opened 2016-03-22
Fixing issues in pre-commit / Travis hooks

mtlynch:
YAPF was complaining about Unicode errors. It seemed to be the result of
the wildcard in the YAPF call, so I've removed it. Also changed the
pre-commit hook slightly to be more consistent with the style of our
other repos.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/66)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,16 +2,16 @@ language: python
 python:
   - "2.7"
 env:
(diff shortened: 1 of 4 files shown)


fernandalavalle:
LGTM

PR #74, opened 2016-03-31
Get slice urls

fernandalavalle:
Added a function to build urls for querying nagios in StatusUpdateHandler.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/74)

### DIFF
--- a/server/mlabns/db/model.py
+++ b/server/mlabns/db/model.py
@@ -185,6 +185,9 @@ def get_all_tool_ids():
     """Gets all Tool model objects.
 
     Returns:
-        Iterable set of Tool instances.
+        List of tool_ids.
     """
-    return Tool.all().run(batch_size=constants.GQL_BATCH_SIZE)
+    tool_ids = []
+    for tool in Tool.all().run(batch_size=constants.GQL_BATCH_SIZE):
+        tool_ids.append(tool.tool_id)
+    return tool_ids
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -331,16 +331,13 @@ class StatusUpdateHandler(webapp.RequestHandler):
 
         nagios_status.authenticate_nagios(nagios)
 
-        for tool in model.get_all_tool_ids():
-            logging.info('Pulling status of %s from Nagios.', tool.tool_id)
-            for family in StatusUpdateHandler.NAGIOS_AF_SUFFIXES:
-                slice_url = nagios.url + '?show_state=1&service_name=' + \
-                      tool.tool_id + family + \
-                      "&plugin_output=1"
+        slices_to_query = nagios_status.get_slice_info(nagios.url,
+                                                       self.NAGIOS_AF_SUFFIXES)
+        for one_slice in slices_to_query:
 
-                slice_status = self.get_slice_status(slice_url)
-                self.update_sliver_tools_status(slice_status, tool.tool_id,
-                                                family)
+            slice_status = self.get_slice_status(one_slice.slice_url)
+            self.update_sliver_tools_status(slice_status, one_slice.tool_id,
+                                            one_slice.ip_version)
         return util.send_success(self)
 
     def update_sliver_tools_status(self, slice_status, tool_id, family):
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -312,11 +312,7 @@ class IPUpdateHandler(webapp.RequestHandler):
 
 
 class StatusUpdateHandler(webapp.RequestHandler):
-    """Updates SliverTools' status from nagios."""
-
-    IPV4 = constants.NAGIOS_IPV4_SUFFIX
-    IPV6 = constants.NAGIOS_IPV6_SUFFIX
-    NAGIOS_AF_SUFFIXES = [IPV4, IPV6]
+    """Updates SliverTools' status from Nagios."""
 
     def get(self):
         """Triggers the update handler.
@@ -331,13 +327,11 @@ class StatusUpdateHandler(webapp.RequestHandler):
 
         nagios_status.authenticate_nagios(nagios)
 
-        slices_to_query = nagios_status.get_slice_info(nagios.url,
-                                                       self.NAGIOS_AF_SUFFIXES)
-        for one_slice in slices_to_query:
+        for slice_info in nagios_status.get_slice_info(nagios.url):
 
-            slice_status = self.get_slice_status(one_slice.slice_url)
-            self.update_sliver_tools_status(slice_status, one_slice.tool_id,
-                                            one_slice.ip_version)
+            slice_status = self.get_slice_status(slice_info.slice_url)
+            self.update_sliver_tools_status(slice_status, slice_info.tool_id,
+                                            slice_info.address_family)
         return util.send_success(self)
 
     def update_sliver_tools_status(self, slice_status, tool_id, family):
@@ -449,7 +443,7 @@ class StatusUpdateHandler(webapp.RequestHandler):
                 sliver_fqdn, state, tool_extra = nagios_status.parse_sliver_tool_status(
                     line)
             except nagios_status.NagiosStatusUnparseableError as e:
-                logging.error('Unable to parse nagios sliver status: %s', e)
+                logging.error('Unable to parse Nagios sliver status: %s', e)
                 continue
 
             if state != constants.NAGIOS_SERVICE_STATUS_OK:
--- a/server/mlabns/tests/test_model.py
+++ b/server/mlabns/tests/test_model.py
@@ -45,11 +45,11 @@ class GetAllToolIdsTest(unittest2.TestCase):
         tool_b.put()
         tool_c.put()
 
-        retrieved_ids = model.get_all_tool_ids()
-        self.assertListEqual(retrieved_ids, expected_ids)
+        actual_ids = model.get_all_tool_ids()
+        self.assertItemsEqual(actual_ids, expected_ids)
 
     def test_get_all_tool_ids_no_stored_tools_returns_empty(self):
-        self.assertListEqual(model.get_all_tool_ids(), [])
+        self.assertItemsEqual(model.get_all_tool_ids(), [])
 
 
 if __name__ == '__main__':
--- a/server/mlabns/tests/test_model.py
+++ b/server/mlabns/tests/test_model.py
@@ -35,31 +35,21 @@ class GetAllToolIdsTest(unittest2.TestCase):
     def tearDown(self):
         self.testbed.deactivate()
 
-    def assertUnorderedToolSetsEqual(self, set_a, set_b):
-        """Equality of sets is based on length and tool_ids of Tools"""
-        list_a = list(set_a)
-        list_b = list(set_b)
-
-        ids_a = [x.tool_id for x in list_a]
-        ids_b = [x.tool_id for x in list_b]
-        self.assertEqual(len(list_a), len(list_b))
-        self.assertListEqual(sorted(ids_a), sorted(ids_b))
-
     def test_get_all_tool_ids_returns_successfully_from_datastore(self):
         tool_a = model.Tool(tool_id='tool_a')
         tool_b = model.Tool(tool_id='tool_b')
         tool_c = model.Tool(tool_id='tool_c')
-        tools = [tool_a, tool_b, tool_c]
+        expected_ids = ['tool_a', 'tool_b', 'tool_c']
 
         tool_a.put()
         tool_b.put()
         tool_c.put()
 
-        retrieved = model.get_all_tool_ids()
-        self.assertUnorderedToolSetsEqual(retrieved, tools)
+        retrieved_ids = model.get_all_tool_ids()
+        self.assertListEqual(retrieved_ids, expected_ids)
 
     def test_get_all_tool_ids_no_stored_tools_returns_empty(self):
-        self.assertUnorderedToolSetsEqual(model.get_all_tool_ids(), [])
+        self.assertListEqual(model.get_all_tool_ids(), [])
 
 
 if __name__ == '__main__':
--- a/server/mlabns/tests/test_nagios_status.py
+++ b/server/mlabns/tests/test_nagios_status.py
@@ -1,6 +1,7 @@
 import mock
 import unittest2
 
+from mlabns.db import model
 from mlabns.util import nagios_status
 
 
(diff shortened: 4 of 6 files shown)


mtlynch:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---

*[server/mlabns/handlers/update.py, line 336 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEAWrYxC0cCvPyw2qAw:-KEAWrYxC0cCvPyw2qAx:-1849901979) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/handlers/update.py#L336)):*
We don't need to save the list. And if we don't need the list, it eliminates the need to distinguish between one slice and many, so suggest:

```
for slice_info in nagios_status.get_slice_info(nagios.url,
                                               self.NAGIOS_AF_SUFFIXES):
```

---

*[server/mlabns/tests/test_nagios_status.py, line 86 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEAhtZ1F9HFsIzVa5GB:-KEAhtZ2hrnRuEqlywMX:1108368988) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/tests/test_nagios_status.py#L86)):*
Any time I see a real URL in a test, my first thought is, "Wait, are they actually calling a production URL in a unit test?" We're not, so suggest making that really obvious with a URL like `http://nagios.mock-mlab.net/baseList`

---

*[server/mlabns/tests/test_nagios_status.py, line 89 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEAc7aWmoeGQzuSxtKy:-KEAc7aWmoeGQzuSxtKz:1694589513) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/tests/test_nagios_status.py#L89)):*
In the other case, we were inheriting from `db.Model` so I suggested a custom equality check so that we don't make assumptions about what equality methods we inherit from `db.Model`.

In this case, `NagiosSliceInfo` is a class we defined from scratch, so in this case, I think we should override `__eq__` and `__ne__` for that class so that we can use `assertListEqual` instead of this custom assert.

---

*[server/mlabns/tests/test_nagios_status.py, line 106 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEAfLfjdkt0q668j1iS:-KEAfLfjdkt0q668j1iT:-2036593721) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/tests/test_nagios_status.py#L106)):*
Always prefer `mock.patch.object` because it can do stronger static checking.

Also, consider doing these in `setUp` to clean up the logic in individual test cases (see example here: https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/tests/test_tool_fetcher.py#L19)

---

*[server/mlabns/tests/test_nagios_status.py, line 108 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEAer6rDoAhMUXcMvi6:-KEAer6rDoAhMUXcMvi7:483539074) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/tests/test_nagios_status.py#L108)):*
This should have occurred to me on the last PR, but if all we'll ever need from `get_tool_ids` is the ID strings, then we should change that function to return just the strings rather than the whole `Tool` instance. Then both functions and tests become much simpler. We can do that in this PR without bloating scope too much.

---

*[server/mlabns/util/nagios_status.py, line 22 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEAXtoSEgwalRcWBwm0:-KEAXtoSEgwalRcWBwm1:1092918292) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/util/nagios_status.py#L22)):*
I think we should mention it's the URL for retrieving status of this particular slice. Also, I think convention in mlab-ns is to uppercase URL.

---

*[server/mlabns/util/nagios_status.py, line 24 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEAX_LFzNFP3CABZiHz:-KEAX_LFzNFP3CABZiI-:-1328153550) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/util/nagios_status.py#L24)):*
Existing convention within mlab-ns (and most of our tools) is to use the term "address family" instead of IP version

---

*[server/mlabns/util/nagios_status.py, line 25 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEAW7R_AQ8ODmBj3qnL:-KEAW7RaHgknxw8-sA0z:536903541) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/util/nagios_status.py#L25)):*
Extraneous blank line

---

*[server/mlabns/util/nagios_status.py, line 32 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEAWUWBatCDtWS0kO2h:-KEAWUWBatCDtWS0kO2i:1813754749) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/util/nagios_status.py#L32)):*
These attributes are all read-only, right? Suggest prefixing them with `_` and adding public`@property` methods to make it explicit:

https://google.github.io/styleguide/pyguide.html?showone=Properties#Properties

---

*[server/mlabns/util/nagios_status.py, line 86 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEAhMg6na1nYMB-sVls:-KEAhMg7UvuaL4BrMT2T:1844346728) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/util/nagios_status.py#L86)):*
Suggest calling it `nagios_base_url` to emphasize that it's not a complete URL

---

*[server/mlabns/util/nagios_status.py, line 87 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEAYuEwkwNqgTnNKWrQ:-KEAYuExWyBXC1lvBz3B:1247490121) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/util/nagios_status.py#L87)):*
It seems like we're inconsistent in this PR about capitalizing Nagios

---

*[server/mlabns/util/nagios_status.py, line 90 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEAanq58bX2U6rU90dt:-KEAanq6-zCIWZjg85mC:-874031650) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/util/nagios_status.py#L90)):*
(nit) missing trailing period

---

*[server/mlabns/util/nagios_status.py, line 92 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEAavhZNR9eognJ_jf-:-KEAavhZNR9eognJ_jf0:-1537362998) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/util/nagios_status.py#L92)):*
Eh "objects" as a name is kind of generic, but I can't really think of anything better. `slice_infos`? Even though "infos" is an awkward pluralization.

---

*[server/mlabns/util/nagios_status.py, line 94 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEA_RvNjZbe_GBt4Ag3:-KEA_RvOJbNZulZRN4Zd:691089735) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/util/nagios_status.py#L94)):*
Why do we need to parameterize the af suffixes? Can we make this like:

```
address_family_suffixes = ['', '_ipv6']
for address_family in address_family_suffixes:
  ...
```

Then we can take AF suffixes out of `constants.py` (which has become kind of a hodgepodge of unrelated constants)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/mlab-ns/74)*

fernandalavalle:
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---

*[server/mlabns/tests/test_nagios_status.py, line 89 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEAc7aWmoeGQzuSxtKy:-KEHeMKjrPmPQ1zCws7b:1371608741) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/tests/test_nagios_status.py#L89)):*
Do you mean override it in the test file? or in the code? Seems weird to change the actual code for tests, right?

---

*[server/mlabns/tests/test_nagios_status.py, line 106 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEAfLfjdkt0q668j1iS:-KEDNGBOILCn_DMwrXnb:-216096582) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/tests/test_nagios_status.py#L106)):*
Ok, so I've actually been working around this for a while. 

So the way I use mock.patch.object is when I want to mock a class or object. Here I'm really trying to get to get_all_tool_ids(). So I just patch the module its in. Is there a more straightforward way of doing that with object? 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/mlab-ns/74)*

mtlynch:
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---

*[server/mlabns/tests/test_nagios_status.py, line 89 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEAc7aWmoeGQzuSxtKy:-KEI0kZNQt675-nSyRmE:-604997333) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/tests/test_nagios_status.py#L89)):*
I think we should override it in the production class. It's reasonable for that class to have `__eq__` and `__ne__` methods. They should be real equality operators though, checking for logical equality on all of the class's public fields.

---

*[server/mlabns/tests/test_nagios_status.py, line 106 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEAfLfjdkt0q668j1iS:-KEI1ctsFGqCvdBFNKgx:819913160) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/tests/test_nagios_status.py#L106)):*
You can do:

```
from mlabns.db import model
...
with mock.patch.object(model, 'get_all_tool_ids') as get_all_tool_ids:
    get_all_tool_ids.return_value = ...
```

In general, I try to avoid using `mock.patch` at all and prefer dependency injection, but we don't need to do that here.

If you're not familiar with dependency injection:

* short version: http://stackoverflow.com/a/130862/3653712
* long version: http://martinfowler.com/articles/injection.html

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/mlab-ns/74)*

fernandalavalle:
Review status: 0 of 6 files reviewed at latest revision, 12 unresolved discussions.

---

*[server/mlabns/util/nagios_status.py, line 92 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEAavhZNR9eognJ_jf-:-KEIbke_zV5pg4LH0cC1:229780225) ([raw file](https://github.com/m-lab/mlab-ns/blob/98fa1f3bca25585fcb6a42f0fb1367362ab239a4/server/mlabns/util/nagios_status.py#L92)):*
slice_info_instances? leaving it for now, because I think slice_infos sounds silly, but if you feel strongly I can update it!

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/mlab-ns/74)*

mtlynch:
:lgtm:

---

Reviewed 6 of 6 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[server/mlabns/tests/test_model.py, line 49 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEIrCxZ37CuozgGvjRG:-KEIrCx_fjepUcJduUKe:967747574) ([raw file](https://github.com/m-lab/mlab-ns/blob/8bbdf25d44896010e266dc173a2e25f69fb33238/server/mlabns/tests/test_model.py#L49)):*
Suggest `assertItemsEqual` to emphasize that order doesn't matter (ditto below)

---

*[server/mlabns/tests/test_nagios_status.py, line 109 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEIs67Wm61iVvW5-nQg:-KEIs67XrOV_c-HDdDaB:-395872832) ([raw file](https://github.com/m-lab/mlab-ns/blob/8bbdf25d44896010e266dc173a2e25f69fb33238/server/mlabns/tests/test_nagios_status.py#L109)):*
Suggest `retrieved` -> `actual` since (expected, actual) are a common pairing in Python

---

*[server/mlabns/util/nagios_status.py, line 38 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-KEIreMTQC8UyF4KgITE:-KEIreMTQC8UyF4KgITF:-1401754848) ([raw file](https://github.com/m-lab/mlab-ns/blob/8bbdf25d44896010e266dc173a2e25f69fb33238/server/mlabns/util/nagios_status.py#L38)):*
Neat, I've never seen this before

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/mlab-ns/74#-:-KEIsvZLJV6QAV8I9InM:1417088721)*

PR #75, opened 2016-03-31
Added a function to retrieve Sliver Tools efficiently in StatusUpdate…

fernandalavalle:
…Handler

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/75)

### DIFF
--- a/server/mlabns/db/model.py
+++ b/server/mlabns/db/model.py
@@ -183,6 +183,7 @@ def get_tool_from_tool_id(tool_id):
 
 def get_all_tool_ids():
     """Gets all Tool model objects.
+
     Returns:
         List of tool_ids.
     """
--- a/server/mlabns/db/model.py
+++ b/server/mlabns/db/model.py
@@ -1,6 +1,5 @@
 import logging
 
-from google.appengine.api import memcache
 from google.appengine.ext import db
 
 from mlabns.util import constants
@@ -189,22 +188,3 @@ def get_all_tool_ids():
         Iterable set of Tool instances.
     """
     return Tool.all().run(batch_size=constants.GQL_BATCH_SIZE)
-
-
-def get_SliverTool_by_tool_id(tool_id):
-    """Retrieves SliverTools by tool_id.
-
-    Returns:
-        List of SliverTool instances on tool_id slice.
-    """
-    sliver_tool_key = 'sliver_tool_tool_id_{}'.format(tool_id)
-    sliver_tools = memcache.get(sliver_tool_key)
-
-    if not sliver_tools:
-        sliver_tools = list(SliverTool.gql(
-            'WHERE tool_id=:tool_id',
-            tool_id=tool_id).run(batch_size=constants.GQL_BATCH_SIZE))
-        if not memcache.set(sliver_tool_key, sliver_tools):
-            logging.error('Failed to update sliver status in memcache.')
-
-    return sliver_tools
--- a/server/mlabns/db/model.py
+++ b/server/mlabns/db/model.py
@@ -183,8 +183,10 @@ def get_tool_from_tool_id(tool_id):
 
 def get_all_tool_ids():
     """Gets all Tool model objects.
-
     Returns:
-        Iterable set of Tool instances.
+        List of tool_ids.
     """
-    return Tool.all().run(batch_size=constants.GQL_BATCH_SIZE)
+    tool_ids = []
+    for tool in Tool.all().run(batch_size=constants.GQL_BATCH_SIZE):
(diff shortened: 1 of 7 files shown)


mtlynch:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[server/mlabns/handlers/update.py, line 354 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/75#-KEAjrSZTmg5Ft19g1Xf:-KEAjrS_ADC4f-hzsB5D:-1211759149) ([raw file](https://github.com/m-lab/mlab-ns/blob/766673789477a3211aa4ac602777538ba4f69c48/server/mlabns/handlers/update.py#L354)):*
You should be able to do this with the existing `ToolFetcher` class without having to write your own custom function:
```
sliver_tools = tool_fetcher.ToolFetcher().fetch(
    tool_fetcher.ToolProperties(tool_id=tool_id))
```

Sorry about the misleading name of `ToolFetcher` since it actually returns `SliverTool`s instead of `Tool`s. Feel free to rename it in this PR if you'd like.

---

*[server/mlabns/tests/test_model.py, line 101 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/75#-KEAnXNAsCTlnvCP-qnF:-KEAnXNBV62D7e5s40kH:-1727922620) ([raw file](https://github.com/m-lab/mlab-ns/blob/766673789477a3211aa4ac602777538ba4f69c48/server/mlabns/tests/test_model.py#L101)):*
(probably moot because I think this code goes away if we use `ToolFetcher` but for future reference): If we get to the point of wanting to check equality on multiple attributes and we control the class, it probably makes sense to override `__eq__` and `__ne__` so we can use standard equality assertions.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/mlab-ns/75)*

mtlynch:
Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/mlabns/handlers/update.py, line 431 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/75#-KEDFYvVBsPYN9Q5GfmX:-KEDFYvWSVr3SgNoKpLn:-1442454614) ([raw file](https://github.com/m-lab/mlab-ns/blob/84a587aae05cbdd7b85ac0074dc4a08dd75164cf/server/mlabns/handlers/update.py#L431)):*
I think this leaves us with a stale memcache because this is the only code that updates tool settings in memcache. We want to be very conservative about preserving functionality here because this is a refactoring PR with a lot of noise from the "sliver" rename.

Suggest renaming the old `sliver_tool_list` to something clearer like `updated_sliver_tools` and keeping this memcache functionality.

Aside: Looking at this code, I bet what's killing us is that `sliver_tool.put()` call on 412 because we're probably doing hundreds of those datastore calls per update, when it could just be a single batched call at the end like we're doing with memcache.

---

*[server/mlabns/tests/test_model.py, line 101 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/75#-KEAnXNAsCTlnvCP-qnF:-KEDEe38TKRU2BcGCEx2:-1681149559) ([raw file](https://github.com/m-lab/mlab-ns/blob/766673789477a3211aa4ac602777538ba4f69c48/server/mlabns/tests/test_model.py#L101)):*
To clarify, I don't think we should have `assertObjectSetsEqualByAttribute`. I think if we find ourselves needing an assertion that's so generic, it's probably time to start thinking about defining equality operators for the objects we're comparing.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/mlab-ns/75)*

mtlynch:
When we squash the commit, make sure we update the commit message so that it doesn't talk about the now-deleted function that retrieves slivers efficiently.

:lgtm:

---

Review status: 0 of 7 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/mlabns/handlers/update.py, line 360 \[r4\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/75#-KEXp3UeRJwTiuWNfpP6:-KEXp3UgOuqexzYTTw9n:1337594546) ([raw file](https://github.com/m-lab/mlab-ns/blob/cb56598c7e3e73fa23feb8ec135c608ae80e6afe/server/mlabns/handlers/update.py#L360)):*
Extraneous whitespace

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/mlab-ns/75#-:-KEXpJ3YtLTzWBPs1lMW:1177833351)*

PR #86, opened 2016-04-20
Added logic to StatusUpdateHandler to deal with blank slice and sliver statuses.

fernandalavalle:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/86)


mtlynch:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 21 unresolved discussions.

---

*[server/mlabns/handlers/update.py, line 335 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoHUckk5--wGuxEbLh:-KFoHUckk5--wGuxEbLi:-222133442) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/handlers/update.py#L335)):*
For this and below, we should use implicit false:

https://google.github.io/styleguide/pyguide.html?showone=True/False_evaluations#True/False_evaluations

---

*[server/mlabns/handlers/update.py, line 338 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoUrs3-_2WZwR2sPKQ:-KFoUrs3-_2WZwR2sPKR:-586135174) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/handlers/update.py#L338)):*
Noooo don't duplicate this magic string in another place to mean IPv4. Why are we special casing IPv4 here?

---

*[server/mlabns/handlers/update.py, line 414 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoINxkRiarNSaz9qg3:-KFoINxlOuydKG56LmEG:-453322103) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/handlers/update.py#L414)):*
Can we pull this function out of `StatusUpdateHandler` entirely? It doesn't use the `self` parameter at all which is a strong sign that it doesn't belong in the class.  Suggest moving it to the `nagios_status` module. 

---

*[server/mlabns/handlers/update.py, line 422 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoJ2A7qGGs7ey2Y-Vl:-KFoJ2A7qGGs7ey2Y-Vm:-1904425767) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/handlers/update.py#L422)):*
Can we update the example return dictionary so that it matches a dictionary this function returns?

---

*[server/mlabns/handlers/update.py, line 433 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoJWWfkGINpk8Quo-H:-KFoJWWfkGINpk8Quo-I:1553126534) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/handlers/update.py#L433)):*
I think it would be easier to do this check before we've split it up. Suggest changing to:

```
nagios_response = lines = urllib2.urlopen(url).read()
...
if not nagios_response or nagios_response.isspace():
...
for line in nagios_response.strip('\n').split('\n'):
...
```

---

*[server/mlabns/handlers/update.py, line 435 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoL15W7sQA0M8AD0qO:-KFoL15W7sQA0M8AD0qP:782906111) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/handlers/update.py#L435)):*
Suggest `Nagios gave empty response for sliver status at the following url:`

---

*[server/mlabns/tests/test_update.py, line 78 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoPLgkts-gZT1Ch_y3:-KFoPLgkts-gZT1Ch_y4:-287872681) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/tests/test_update.py#L78)):*
Suggest renaming to `mock_urlopen_response`. I had a really hard time following how it fit in.

---

*[server/mlabns/tests/test_update.py, line 96 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoM6kQgb_8MC-eDM9A:-KFoM6kQgb_8MC-eDM9B:-259427128) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/tests/test_update.py#L96)):*
Why are we patching this? It doesn't look like we use it. We generally shouldn't be testing logging unless the purpose of the thing we're testing is logging. Otherwise logging is kind of an implementation detail and we shouldn't test those.

---

*[server/mlabns/tests/test_update.py, line 100 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoNzJYEVdgfep92mrd:-KFoNzJYEVdgfep92mre:720938050) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/tests/test_update.py#L100)):*
Suggest clarifying the comment for this one and the one below. Is this an expected return value from something? It seems like this is a dummy *input* to something, so not really something that we expect back.

---

*[server/mlabns/tests/test_update.py, line 101 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoVAHoQzoSlb_UiWr_:-KFoVAHoQzoSlb_UiWra:-701117989) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/tests/test_update.py#L101)):*
Suggest `site1` -> `xyz01`. If we're going to kind of match the schema of M-Lab URLs, we should match it in a recognizable way.

---

*[server/mlabns/tests/test_update.py, line 104 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoO34ib5gTgoZEpZ78:-KFoMUphQTN8Y-8-mRA6:1393856447) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/tests/test_update.py#L104)):*
Suggest defining this in one call with a triple-quoted multi-line string.

---

*[server/mlabns/tests/test_update.py, line 116 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoNmZw1Hs3QtDL-kHM:-KFoNmZw1Hs3QtDL-kHN:-2052759265) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/tests/test_update.py#L116)):*
`mock_slice_status` and `parsed_statuses` are both constants and don't depend on `self` so they should go outside the class like (`MOCK_SLICE_STATUS` and `PARSED_STATUSES`)

Edit: Actually, after reading the rest of the tests, I'd say we should define them inline in each test where they're needed instead of sharing them at the module level. There are times when we want to define stuff at the module level, but they come at the cost of making the reader jump around our file to understand what's going on, so we should really only do it if it saves us a lot of repeated code.

As is, we really only use these two variables in this form in one function (the success test), then in the other two cases, we're kind of bending over backwards to make them work when we could probably be better off redefining fresh. See my note a few down about using a static list for the side effect instead of a dictionary.

Also, suggest reducing it to 3 sites because we're not really getting much additional value from 4 vs. 3 and it's just more code.

---

*[server/mlabns/tests/test_update.py, line 120 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoOcCbK80iBTwm_Oa3:-KFoOcCbK80iBTwm_Oa4:-1243189468) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/tests/test_update.py#L120)):*
Functions should never begin with a blank line (this repeats a few functions down)

---

*[server/mlabns/tests/test_update.py, line 123 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoOrr9kW-0iZXL44XW:-KFoOrr9kW-0iZXL44XX:1414814828) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/tests/test_update.py#L123)):*
Suggest ditching the extra variable and assigning the `lambda` directly to the `side_effect` on the next line.

---

*[server/mlabns/tests/test_update.py, line 124 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoQ5CRCvoFmK2IoSes:-KFoQ5CRCvoFmK2IoSet:-1506493440) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/tests/test_update.py#L124)):*
Actually, this is pretty confusing. I think it's easier to have it return a static list of return values and we assume the ordering. It's arguably more brittle because it assumes an implementation detail (namely that `get_slice_status` will parse lines in the order it gets them from Nagios), but I think it's worth the reduction in indirection.

So it'd be like:

```
update.nagios_status.parse_sliver_tool_status.side_effect = [
  ('mock.mlab1.site1.measurement-lab.org', '0', 'mock tool extra'), 
  ('mock.mlab2.site1.measurement-lab.org', '0', 'mock tool extra'),
  ('mock.mlab3.site1.measurement-lab.org', '0', 'mock tool extra'),
  ('mock.mlab1.site2.measurement-lab.org', '0', 'mock tool extra')
  ]
```

Ditto for the rest of the test cases.

---

*[server/mlabns/tests/test_update.py, line 147 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoPZnJS34u96OWpgxv:-KFoPZnJS34u96OWpgxw:822400651) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/tests/test_update.py#L147)):*
We should also `assert_called_once_with` to make sure we're calling `urlopen` with the correct URL.

---

*[server/mlabns/tests/test_update.py, line 149 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoVyWWeXyNwCaKzsf9:-KFoVyWWeXyNwCaKzsfA:-497428365) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/tests/test_update.py#L149)):*
Suggest rephrasing to make these read more like sentences, so like `test_get_slice_status_returns_None_when_nagios_response_is_whitespace_and_no_newline`

(ditto for other tests)

---

*[server/mlabns/tests/test_update.py, line 164 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoVfEy5v2Dda1Wtw3v:-KFoVfEy5v2Dda1Wtw3w:856756377) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/tests/test_update.py#L164)):*
I think we should actually preserve capitalization for class names

---

*[server/mlabns/tests/test_update.py, line 190 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoUA2mNlNzDkcCv9ST:-KFoTpeefHZXfQHKpqqy:-1399033733) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/tests/test_update.py#L190)):*
Can we use the structure:

```
with self.assertRaises(nagios_status.NagiosStatusUnparseableError):
  update.nagios_status.parse_sliver_tool_status.(
            'mock.mlab2.site1.measurement-lab.org/ndt 0 1 mock tool extra')
```

I think it's way more obvious. (ditto for the rest of the file)

Edit: Wait, we don't have to assert this. We defined this behavior a few lines up. This is like a test of our test, so we should get rid of it.

---

*[server/mlabns/tests/test_update.py, line 235 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoLdoiIvFrNP7HtUi1:-KFoLdoiIvFrNP7HtUi2:1138072844) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/tests/test_update.py#L235)):*
Do we actually need to define our own error class? I think we should be able to do `urllib2.HTTPError('mock http error')`. If not, suggest defining the custom error class within this test, since it's the only test that needs it.

---

*[server/mlabns/tests/test_update.py, line 240 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoQM1WDTBuSGSKAZ1O:-KFoQM1XMNYaxlq8s-wD:-2039927001) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/tests/test_update.py#L240)):*
I think this is more a test of the nagios parser, which is out of scope of this test class. Suggest ditching this function and instead adding more diversity to the success input/output (e.g. make one of them offline, make one of them have no tool extra field)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/mlab-ns/86)*

fernandalavalle:
Review status: 0 of 4 files reviewed at latest revision, 21 unresolved discussions.

---

*[server/mlabns/handlers/update.py, line 338 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoUrs3-_2WZwR2sPKQ:-KFou5BiQatSYkhRGo69:231803702) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/handlers/update.py#L338)):*
Okay, yes I know. I thought about how this was digging a deeper hole but from what we talked about evaluating whether a service name is ipv4 or ipv6 is something that should be native to NagiosSliceInfo which this would have access to, so I thought it wasn't too terrible. 

I added this check because from what I understood we don't expect blank statuses for any service on ipv4 but we expect several services to always return blank slice statuses for ipv6. I didn't want the ipv4 to fail silently. 

---

*[server/mlabns/handlers/update.py, line 433 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoJWWfkGINpk8Quo-H:-KFt8xGDfpeWtUg9CEw8:-140720236) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/handlers/update.py#L433)):*
So that's not totally the same. What you suggested just checks for blank slice statuses (the whole thing), and the current solution checks for blank sliver statuses (each line) too. 

You think its reasonable to assume that if the slice status isn't empty or blank, we won't get blank lines for sliver statuses? Seems fine to me, I just want to check.

---

*[server/mlabns/tests/test_update.py, line 104 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoO34ib5gTgoZEpZ78:-KFtFJmuBJfgSATGN-Fm:391671373) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/tests/test_update.py#L104)):*
The indentations mess up parsing. Do you still prefer the triple-quote with awkward dedent?

---

*[server/mlabns/tests/test_update.py, line 124 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoQ5CRCvoFmK2IoSes:-KFtaL-F7JPtG_LD4lVX:429485859) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/tests/test_update.py#L124)):*
You can't use this format for the tests that need parse_sliver_tool_status to raise exceptions, can you? 

Is self consistency less valuable than less indirection? Should I only implement this for the tests currently using lambda?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/mlab-ns/86)*

mtlynch:
Review status: 0 of 4 files reviewed at latest revision, 22 unresolved discussions.

---

*[server/mlabns/handlers/update.py, line 338 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoUrs3-_2WZwR2sPKQ:-KFuGcrmoZVXY1TSuBzA:-593037056) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/handlers/update.py#L338)):*
I think we should just cut this check. I don't want to bake assumptions into mlab-ns that missing IPv4 is unusual but missing IPv6 is not. That's something that just happens to be true today, but it feels like that's pushing too much knowledge of our current state into mlab-ns in a not-so-obvious place.

Plus, really anything to stop us from spreading that magic string.

---

*[server/mlabns/handlers/update.py, line 414 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoINxkRiarNSaz9qg3:-KFuHrtddRNifslTsoWc:1115866111) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/handlers/update.py#L414)):*
Looks like you moved it but forgot to delete the original

---

*[server/mlabns/handlers/update.py, line 433 \[r1\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFoJWWfkGINpk8Quo-H:-KFuID2AIodNzsdMqTRF:-191168895) ([raw file](https://github.com/m-lab/mlab-ns/blob/3c1b247cb28bd5e6e7a5d0e878fe66743fb980b6/server/mlabns/handlers/update.py#L433)):*
Oh right. I forgot about blank lines mixed in among non-blank lines. Sorry, your original implementation was right. 

Can you add a comment to say "Remove any blank lines" and a unit test for the case of blank lines mixed with non-blanks?

---

*[server/mlabns/tests/test_nagios_status.py, line 147 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFuK-WYYSFcybUjwqz8:-KFuK-WZoD68UEvQM2h6:-1475549640) ([raw file](https://github.com/m-lab/mlab-ns/blob/b31610a52033f221ac4a02e811f3d8417cdca5bb/server/mlabns/tests/test_nagios_status.py#L147)):*
>The indentations mess up parsing. Do you still prefer the triple-quote with awkward dedent?

Suggest either:

```
mock_slice_status = """
mock.mlab1.xyz01.measurement-lab.org/ndt 0 1 mock tool extra
mock.mlab2.xyz01.measurement-lab.org/ndt 0 1 mock tool extra
mock.mlab3.xyz01.measurement-lab.org/ndt 2 1 mock tool extra
""".lstrip()
```

or

```
mock_slice_status = """mock.mlab1.xyz01.measurement-lab.org/ndt 0 1 mock tool extra
mock.mlab2.xyz01.measurement-lab.org/ndt 0 1 mock tool extra
mock.mlab3.xyz01.measurement-lab.org/ndt 2 1 mock tool extra
"""
```

---

*[server/mlabns/tests/test_nagios_status.py, line 150 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFuKj2WW4TQ3SLiEES3:-KFuKj2X_UlnL0nktmAR:1123428453) ([raw file](https://github.com/m-lab/mlab-ns/blob/b31610a52033f221ac4a02e811f3d8417cdca5bb/server/mlabns/tests/test_nagios_status.py#L150)):*
We don't have to `iter`. `side_effect` knows that if we give it a list, it should iterate through it for return values.

---

*[server/mlabns/tests/test_nagios_status.py, line 156 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFuJJAEI-tIpAEaLYtw:-KFuJJAEI-tIpAEaLYtx:324791740) ([raw file](https://github.com/m-lab/mlab-ns/blob/b31610a52033f221ac4a02e811f3d8417cdca5bb/server/mlabns/tests/test_nagios_status.py#L156)):*
>You can't use this format for the tests that need parse_sliver_tool_status to raise exceptions, can you?

You actually can. `side_effect` has convenient but not intuitive semantics where if there are exceptions in the list, it raises them instead of returning them, so like:

```
mock_whatever.foo.side_effect = [1, ValueError('bar'), 2]

mock_whatever.foo() # returns 1
mock_whatever.foo() # raises ValueError
mock_whatever.foo() # returns 2

```

>Is self consistency less valuable than less indirection? Should I only implement this for the tests currently using lambda?

I'm not sure if this is moot from the first answer, but I think we should define it inline each time since I don't think we're saving ourselves much code by reusing it.

---

*[server/mlabns/tests/test_nagios_status.py, line 175 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFuLC3dDAUbyKxCUcej:-KFuLC3dDAUbyKxCUcek:-1389953821) ([raw file](https://github.com/m-lab/mlab-ns/blob/b31610a52033f221ac4a02e811f3d8417cdca5bb/server/mlabns/tests/test_nagios_status.py#L175)):*
`assertDictEqual`

---

*[server/mlabns/tests/test_nagios_status.py, line 180 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFuLLeSJ_Y667DIPzTI:-KFuLLeSJ_Y667DIPzTJ:1534733149) ([raw file](https://github.com/m-lab/mlab-ns/blob/b31610a52033f221ac4a02e811f3d8417cdca5bb/server/mlabns/tests/test_nagios_status.py#L180)):*
Suggest assigning the string directly instead of going through a variable for one line (ditto for the next test as well). There's some value in giving the string a human-friendly name, but I think it's obvious enough from the test name and the string itself.

---

*[server/mlabns/util/nagios_status.py, line 138 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFuIzH3oWtjAF6fO2EO:-KFuIzH3oWtjAF6fO2EP:1757939127) ([raw file](https://github.com/m-lab/mlab-ns/blob/b31610a52033f221ac4a02e811f3d8417cdca5bb/server/mlabns/util/nagios_status.py#L138)):*
Suggest replacing the constant (because the caller gets back a string not a symbolic constant) and reformatting to

````
{
   'foo.mlab1.site1.measurement-lab.org': {
        'status': 'online',
        'tool_extra': 'example tool extra'
        }
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/mlab-ns/86)*
<!-- Sent from Reviewable.io -->

mtlynch:
Reviewed 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 22 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/mlab-ns/86)*

mtlynch:
Review status: all files reviewed at latest revision, 22 unresolved discussions.

---

*[server/mlabns/tests/test_nagios_status.py, line 147 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFuK-WYYSFcybUjwqz8:-KFuOH9x-YcyKaLvBZ26:-674684427) ([raw file](https://github.com/m-lab/mlab-ns/blob/b31610a52033f221ac4a02e811f3d8417cdca5bb/server/mlabns/tests/test_nagios_status.py#L147)):*
Oh I maybe didn't answer your question. Yeah, I think a dedent is fine.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/mlab-ns/86)*

fernandalavalle:
Review status: 1 of 4 files reviewed at latest revision, 20 unresolved discussions.

---

*[server/mlabns/tests/test_nagios_status.py, line 150 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFuKj2WW4TQ3SLiEES3:-KG0PeaD_Dpwns3KaAtq:-1857636224) ([raw file](https://github.com/m-lab/mlab-ns/blob/b31610a52033f221ac4a02e811f3d8417cdca5bb/server/mlabns/tests/test_nagios_status.py#L150)):*
I don't know it throws an error when I don't make it an iterator. Are you sure that's unexpected behavior? Am I doing something weird?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/mlab-ns/86)*

mtlynch:
Reviewed 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 18 unresolved discussions.

---

*[server/mlabns/handlers/update.py, line 334 \[r3\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KGDras3M6wmZJE_fZf0:-KGDras3M6wmZJE_fZf1:-483840076) ([raw file](https://github.com/m-lab/mlab-ns/blob/3501551b4621efe07bb402b2850cd5712407933a/server/mlabns/handlers/update.py#L334)):*
I think this has to become `nagios_status.get_slice_status`

---

*[server/mlabns/tests/test_nagios_status.py, line 150 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFuKj2WW4TQ3SLiEES3:-KGDr6muB0qWt50tOXh1:1653930822) ([raw file](https://github.com/m-lab/mlab-ns/blob/b31610a52033f221ac4a02e811f3d8417cdca5bb/server/mlabns/tests/test_nagios_status.py#L150)):*
Likely so! Here's a working example:

https://github.com/fernandalavalle/mlab-ns/compare/less_logging_and_blank_status...mtlynch:less_logging?expand=1

---

*[server/mlabns/tests/test_nagios_status.py, line 220 \[r3\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KGDoy8b54wXd2aF7dVx:-KGDoy8b54wXd2aF7dVy:1120673807) ([raw file](https://github.com/m-lab/mlab-ns/blob/3501551b4621efe07bb402b2850cd5712407933a/server/mlabns/tests/test_nagios_status.py#L220)):*
We don't need to assign to a temp variable here, we can assign directly to `return_value`

---

*[server/mlabns/util/nagios_status.py, line 138 \[r2\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KFuIzH3oWtjAF6fO2EO:-KGDpAQCgmwCjTxwOHRU:-1399480240) ([raw file](https://github.com/m-lab/mlab-ns/blob/b31610a52033f221ac4a02e811f3d8417cdca5bb/server/mlabns/util/nagios_status.py#L138)):*
ping on reformatting

---

*[server/mlabns/util/nagios_status.py, line 163 \[r3\]](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-KGDs_4yYItYObY0AUMk:-KGDs_4zA5PSC3967Szj:-2013382548) ([raw file](https://github.com/m-lab/mlab-ns/blob/3501551b4621efe07bb402b2850cd5712407933a/server/mlabns/util/nagios_status.py#L163)):*
Negation is harder for humans to parse, so suggest flipping the order to avoid it:

```
if state == constants.NAGIOS_SERVICE_STATUS_OK:
  status[sliver_fqdn] = {
    'status': message.STATUS_ONLINE,
    'tool_extra': tool_extra
    }
else:
  status[sliver_fqdn] = {
    'status': message.STATUS_OFFLINE,
    'tool_extra': tool_extra
    }
```

(optional) Also, if we want to, we can reduce lines like:

```
status[sliver_fqdn] = { 'tool_extra': tool_extra }
if state == constants.NAGIOS_SERVICE_STATUS_OK:
  status[sliver_fqdn]['status'] = message.STATUS_ONLINE
else:
  status[sliver_fqdn]['status'] = message.STATUS_OFFLINE

```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/mlab-ns/86)*

mtlynch:
:lgtm:, just fix the mock dependency so Travis builds

---

Reviewed 3 of 3 files at r4.
Review status: all files reviewed at latest revision, 18 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/mlab-ns/86#-:-KGIZlwHNfpePDsaU00Z:541796652)*

Issue #38, opened 2015-07-16
NPAD is failing to perform redirects


mtlynch:
@nkinkade reported that mlab-ns is failing to redirect for NPAD.

http://mlab-ns.appspot.com/npad?format=redirect

Seems to be a missed unicode->str conversion someplace.
```
Traceback (most recent call last):
  File "/base/data/home/runtimes/python27/python27_lib/versions/third_party/webapp2-2.5.1/webapp2.py", line 1547, in __call__
    return response(environ, start_response)
  File "/base/data/home/runtimes/python27/python27_lib/versions/third_party/webob-1.1.1/webob/response.py", line 939, in __call__
    start_response(self.status, headerlist)
  File "/base/data/home/runtimes/python27/python27_lib/versions/1/google/appengine/ext/appstats/recording.py", line 1284, in appstats_start_response
    return start_response(status, headers, exc_info)
  File "/base/data/home/runtimes/python27/python27_lib/versions/1/google/appengine/runtime/wsgi.py", line 200, in _StartResponse
    (_GetTypeName(value), value, name))
InvalidResponseError: header values must be str, got 'unicode' (u'http://npad.iupui.mlab2.nuq04.measurement-lab.org:8000') for 'Location'
```

critzo:
Bumping this thread since it was just reported again by a user who emailed support. @mtlynch @fernandalavalle

mtlynch:
Bumping because we got an additional report of a user running into this

critzo:
We have received another report of board failing to run because of this bug. @fernandalavalle

PR #92, opened 2016-08-09
Npad redirect

fernandalavalle:
The way mlab-ns handles npad?format=redirect http requests is currently broken. If a request does not specify a port, the request is answered with a 404. This short term solution specifies port  8000 for all npad redirect requests. This is a constant port that should not need to be changed.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/92)


fernandalavalle:
Addresses: 
https://github.com/m-lab/mlab-ns/issues/38

server/mlabns/handlers/lookup.py
```diff
@@ -23,7 +23,7 @@ def _create_tool_url(fqdn, http_port):
     Returns:
         A constructed tool HTTP URL with protocol and http port.
     """
-    return 'http://%s:%s' % (fqdn, http_port)
+    return str('http://%s:%s' % (fqdn, http_port))
```
pboothe: This change looks redundant and like it does nothing?  Did you mean:

`return 'http://%s:%s' % (fqdn, str(http_port))`

or maybe

`return 'http://%s:%d' % (fqdn, http_port)`

?

fernandalavalle: Yeah, just looking at it I agree it looks redundant. If I take it out though, I get the unicode error we see on mlab-ns now (http://mlab-ns.appspot.com/npad?format=redirect). So I definitely have to cast something as a string, but I'm open to more clever/logical approaches.

pboothe: Put a comment above it saying
  `# Use str() to ensure we return a non-unicode string` 
and I am happy.

pboothe:
LGTM once the comment is added.

PR #93, opened 2017-03-24
Maxmind db update

fernandalavalle:
The maxmind database got separated into separate IPv4 and IPv6 files. 
This PR adds logic to mlab-ns to check the format of the look up ip to point to the right maxmind database.


server/mlabns/util/maxmind.py
```diff
@@ -29,16 +30,24 @@ def __ne__(self, other):
...
     Returns:
         A populated GeoRecord if matching geolocation data is found for the
         IP address. Otherwise, an empty GeoRecord.
     """
+    maxmind_file = None
+    if address_family == message.ADDRESS_FAMILY_IPv4:
+        maxmind_file = constants.GEOLOCATION_MAXMIND_CITY_FILE_IPv4
+    elif address_family == message.ADDRESS_FAMILY_IPv6:
+        maxmind_file = constants.GEOLOCATION_MAXMIND_CITY_FILE_IPv6
+
```
collina: Should we throw something here if address family isn't properly set, or reduce to if/else? Seems weird to have if/elif when it needs to be one of the two to function.

server/mlabns/util/lookup_query.py
```diff
@@ -193,8 +197,17 @@ def _get_user_defined_lat_lon(self, request):
...
+        if _is_valid_ipv6(ip_address):
+            address_family = message.ADDRESS_FAMILY_IPv6
+        elif _is_valid_ipv4(ip_address):
+            address_family = message.ADDRESS_FAMILY_IPv4
+
         if ip_address is not None:
-            geo_record = maxmind.get_ip_geolocation(ip_address)
+            logging.debug('Getting maxmind info for ip %s in family %s',
+                          ip_address, address_family)
+            geo_record = maxmind.get_ip_geolocation(ip_address, address_family)
```
collina: Do we need to do the v4/v6 logic in this function and pass the results in a parameter? Seems like it would be more simple to do it in `get_ip_geolocation` -- are you doing this because you already have `_is_valid_ipv6` here?

server/mlabns/util/maxmind.py
fernandalavalle: Well I wasn't sure. The ip should be properly formatted because that's checked when the LookupQuery object is created. My thought process for not having an else is just that I don't like sneaky errors, if it's not set to ipv4 or ipv6 here it should remain None which would be a more obvious error. 

I could be persuaded to change to if/else though if you think it's better.

server/mlabns/util/lookup_query.py
fernandalavalle: Yeah, I figured there'd be a little redundancy either for functions checking the family or for the logic. I'm ambivalent about which. Which do you prefer?

server/mlabns/util/maxmind.py
pboothe: `if / elif / else log an error and proceed with the error case` is the right way to do things when you are working in a domain that has two right answers and many many wrong ones.

PR #96, opened 2017-04-28
Fixes a bug where not all sliver tools got added to memcache or DS on update

nkinkade:
This PR does two things, hopefully.

### Thing 1
There was a bug in the logic of my previous commit which would allow Nagios updates to propagate to only one of the `ndt` and `ndt_ssl` "tools". As was [noted by Michael in a PR a long time ago](https://github.com/m-lab/mlab-ns/pull/53), we now have a situation where a single FQDN is not unique to a given "tool". This is true in the case of the `ndt` and `ndt_ssl` tools (e.g., Nagios service checks).

There is currently [a `for` loop going over all sliver tools returned for a given FQDN](https://github.com/m-lab/mlab-ns/blob/a609ecce2869afc41e53de21fed38ef8b8b5aa70/server/mlabns/handlers/update.py#L235). The problem is that only the last of the sliver_tools will actually make it into memcache. The value of sliver_tool in the `for` loop would be overwritten by the one in the next iteration, and only the last one would eventually make it down to the [point where it was added to the list sliver_tools to update in the datastore and memcache](https://github.com/m-lab/mlab-ns/blob/a609ecce2869afc41e53de21fed38ef8b8b5aa70/server/mlabns/handlers/update.py#L263).

This also makes sure that both `ndt` and `ndt_ssl` entries exist in the datastore for every FQDN in the `ndt` tool, whereas before the only time that entries for the `ndt_ssl` tool ever got written to the datastore was when a new site showed up in _mlab-site-stats.json_. If an entry got deleted in some way later, it would never show up in the datastore again.

### Thing 2
This one is related to fixing Thing1. Before, we were looping on sliver_tool, and now we loop on tool_id, since this allows us to simplify the flow. We can now query the datastore for a unique record (on both `fqdn` and `tool_id`), so that we can process each sliver_tool one at a time.  Previously, we had one half of an `if` block looping on multiple sliver_tools, and the other half working on a single (presently non-existent) sliver_tool, yet there were operations that needed to be performed on all sliver_tools, both old and new.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/96)

### DIFF
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -225,33 +225,36 @@ class IPUpdateHandler(webapp.RequestHandler):
             # a given slice_id might have multiple tools (e.g., iupui_ndt has
             # both 'ndt' and 'ndt_ssl' tools.
             tools = model.Tool.gql('WHERE slice_id=:slice_id',
-                                  slice_id=slice_id)
+                                   slice_id=slice_id)
             if tools.count() == 0:
                 continue
 
             for tool in tools.run():
                 # Query the datastore to see if this sliver_tool exists there.
                 sliver_tool_gql = model.SliverTool.gql(
-                        'WHERE fqdn=:fqdn AND tool_id=:tool_id',
-                        fqdn=fqdn,
-                        tool_id=tool.tool_id)
+                    'WHERE fqdn=:fqdn AND tool_id=:tool_id',
+                    fqdn=fqdn,
+                    tool_id=tool.tool_id)
 
                 # Check to see if the sliver_tool already exists in the
                 # datastore. If not, add it to the datastore.
                 if sliver_tool_gql.count() == 1:
                     sliver_tool = sliver_tool_gql.get(
-                            batch_size=constants.GQL_BATCH_SIZE)
+                        batch_size=constants.GQL_BATCH_SIZE)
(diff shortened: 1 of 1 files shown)


pboothe:
LGTM

server/mlabns/handlers/update.py
```diff
@@ -214,57 +214,59 @@ def get(self):
...
+                    tool_id=tool.tool_id)
+
+                # Check to see if the sliver_tool already exists in the
+                # datastore. If not, add it to the datastore.
+                if sliver_tool_gql.count() == 1:
+                    sliver_tool = sliver_tool_gql.get(
+                        batch_size=constants.GQL_BATCH_SIZE)
+                elif sliver_tool_gql.count() == 0:
+                    logging.info(
+                        'For tool {}, fqdn {} is not in datastore.  Adding it.'.format(
```
pboothe: prefer 
```python
logging.info('For tool %s, fqdn %s is not in datastore.  Adding it.', tool.tool_id, fqdn)
```
because then if the log level is such that INFO messages are not printed, then the string formatting work will not be done. This is a tiny optimization, and doesn't matter here, so LGTM either way.

nkinkade: Implemented. Thanks!

PR #102, opened 2017-08-07
Sets a slivertool to offline if it exists in memcache, but not in Nagios.

nkinkade:
A small bug in mlab-ns was uncovered by a mistake I made when [decommissioning several sites](https://github.com/m-lab/operator/pull/143). I removed the sites from sites.py and Nagios without first having put them in lame-duck mode. The end result was that memcache in mlab-ns had correctly had the entries for these sites removed, but the datastore still had records for the removed sites, and worse had them marked as "online".

The problem was that when a query comes into mlab-ns and it cannot find a result in memcache, it will fall back to looking for something in the datastore. And in this case, the datastore had old/bad data for a site that no longer existed.

This problem would have been mitigated by my having put the nodes at these sites into lame-duck mode before removing them from Nagios and sites.py. The old entries in the datastore would still have been there, but they would have at least been marked as offline, and thus benign.

This PR implements yet one further safeguard against this sort of problem happening again. It adds some logic wherein if the routine (run once a minute) finds that it knows about a slivertool that for some reason is now not showing up in Nagios (baseList.pl), it will mark that slivertool as offline.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/102)

### DIFF
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -377,44 +377,45 @@ class StatusUpdateHandler(webapp.RequestHandler):
         updated_sliver_tools = []
         for sliver_tool in sliver_tools:
 
-            if sliver_tool.fqdn not in slice_status:
-                logging.info('Nagios does not know sliver %s.',
-                             sliver_tool.fqdn)
-                continue
-
-            if family == '':
-                if sliver_tool.sliver_ipv4 == message.NO_IP_ADDRESS:
-                    if sliver_tool.status_ipv4 == message.STATUS_ONLINE:
-                        logging.warning('Setting IPv4 status of %s to offline '\
-                                        'due to missing IP.', sliver_tool.fqdn)
-                        sliver_tool.status_ipv4 = message.STATUS_OFFLINE
+            if sliver_tool.fqdn in slice_status:
+                if family == '':
+                    if sliver_tool.sliver_ipv4 == message.NO_IP_ADDRESS:
+                        if sliver_tool.status_ipv4 == message.STATUS_ONLINE:
+                            logging.warning('Setting IPv4 status of %s to offline '\
+                                            'due to missing IP.', sliver_tool.fqdn)
+                            sliver_tool.status_ipv4 = message.STATUS_OFFLINE
+                    else:
+                        if (sliver_tool.status_ipv4 !=
+                                slice_status[sliver_tool.fqdn]['status'] or
+                                sliver_tool.tool_extra !=
+                                slice_status[sliver_tool.fqdn]['tool_extra']):
+                            sliver_tool.status_ipv4 = \
+                              slice_status[sliver_tool.fqdn]['status']
+                            sliver_tool.tool_extra = \
+                              slice_status[sliver_tool.fqdn]['tool_extra']
+                elif family == '_ipv6':
+                    if sliver_tool.sliver_ipv6 == message.NO_IP_ADDRESS:
+                        if sliver_tool.status_ipv6 == message.STATUS_ONLINE:
+                            logging.warning('Setting IPv6 status for %s to offline'\
+                                            ' due to missing IP.', sliver_tool.fqdn)
+                            sliver_tool.status_ipv6 = message.STATUS_OFFLINE
+                    else:
+                        if (sliver_tool.status_ipv6 !=
+                                slice_status[sliver_tool.fqdn]['status'] or
+                                sliver_tool.tool_extra !=
+                                slice_status[sliver_tool.fqdn]['tool_extra']):
+                            sliver_tool.status_ipv6 = \
+                              slice_status[sliver_tool.fqdn]['status']
+                            sliver_tool.tool_extra = \
+                              slice_status[sliver_tool.fqdn]['tool_extra']
                 else:
-                    if (sliver_tool.status_ipv4 !=
-                            slice_status[sliver_tool.fqdn]['status'] or
-                            sliver_tool.tool_extra !=
-                            slice_status[sliver_tool.fqdn]['tool_extra']):
-                        sliver_tool.status_ipv4 = \
-                          slice_status[sliver_tool.fqdn]['status']
-                        sliver_tool.tool_extra = \
-                          slice_status[sliver_tool.fqdn]['tool_extra']
-            elif family == '_ipv6':
-                if sliver_tool.sliver_ipv6 == message.NO_IP_ADDRESS:
-                    if sliver_tool.status_ipv6 == message.STATUS_ONLINE:
-                        logging.warning('Setting IPv6 status for %s to offline'\
-                                        ' due to missing IP.', sliver_tool.fqdn)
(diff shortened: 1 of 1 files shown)


pboothe:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/mlab-ns/102#-:-KqyaVLwXlZq7sSbrHFf:bnfp4nl)*

pboothe:
At the request of @stephen-soltesz, I just double-checked that if Nagios is down this won't mark everything as offline.  And I think that the fact that it is called in a loop as:
```python
        for slice_info in nagios_status.get_slice_info(nagios.url):

            slice_status = nagios_status.get_slice_status(slice_info.slice_url)
            if slice_status:
                self.update_sliver_tools_status(
                    slice_status, slice_info.tool_id, slice_info.address_family)
```

means that if `nagios_status.get_slice_info` fails, then this function won't get called. I grepped through the codebase and found no other calls to `update_sliver_tools_status`

nkinkade:
@stephen-soltesz, @pboothe: To slightly extend what Peter found, [`nagios_status.get_slice_status()` will return `None`](https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/util/nagios_status.py#L143) if there are no rows returned from Nagios or if the URL can't be reached. The remaining question is whether **any** valid output from Nagios is good enough? Put another way, is there a case where baseList.pl could return a subset of valid output, but not all of it?

**Update**: I wrote the above just before we chatted on the mlab VC about this, and it sounds like the answer to the question above is "maybe. we don't know." In that case, the agreement is to add some sort of sanity checking on the data received from baseList.pl to be *reasonably* sure that we got everything and not just partial/truncated data. 

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/mlab-ns/102#-:-KqyjksM-AaHjkptAin9:bq7ioj1)*

nkinkade:
@stephen-soltesz, @pboothe: I've done a bit of testing with urllib2 to see how it reacts to various network issues:

* In the case of an established TCP connection pretty much vanishing (iptables blocking bidirectional traffic), urllib2 seems to hang indefinitely. It would appear that the underlying socket module's default for a timeout is `None`, which seems to mean no timeout at all.

* In the case of an established TCP connection suddenly receiving RSTs from the remote end, urllib2 eventually throws a `<urlopen error [Errno 101] Network is unreachable>` error.

* In the case where an established TCP connection is closed gracefully by the server, but **before** all of the request content has been transferred, urllib2 does not throw an exception (even when the `Content-Length` header is provided`) and will deliver the partial content on `read()`. It seems that this is the possible case we'd need to be concerned about.

nkinkade:
For the sake of reproducibility, if anyone wanted to wanted to recreate the tests I did. On some server I did:

```
$ truncate -s 100M rofl.bin
$ echo -e "\nROFL" >> rofl.bin
$ nc -l -p 9999 < <(echo -ne "HTTP/1.1 200 OK\nContent-Length: 104857606\n\n"; cat rofl.bin)
```

Then on the client I had this tiny python script:

```
#!/usr/bin/python

import urllib2
try:
    response = urllib2.urlopen('http://natha.nkinka.de:9999/')
except Exception, e:
    print "ERROR: ", e
html = response.read()
open('rofl.out', 'w').write(html)
```

Then on the server I would implement firewall rules to either DROP or REJECT (--reject-with tcp-reset) traffic to TCP port 9999. And to test a graceful TCP closure (but before the transfer was complete), I would just ctrl-c the netcat process.

stephen-soltesz:
So, as a result of this change, it's sounding like the state of mlab-ns is more exposed to misconfigurations in the Nagios server than before. I'm satisfied that the network level failures won't affect mlab-ns negatively.

How comfortable are we depending on the correctness of the Nagios configs? My thought is that as long as we're conscious of this dependency and exposure (and are mindful of config updates), then it is okay.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/mlab-ns/102#-:-Kr2JXzDODyYNAwdR7kI:bxfznka)*

nkinkade:
I'm very comfortable counting on the correctness of the Nagios status log (which baseList.pl reads), and I'm very comfortable with the reliability of baseList.pl. What now gives me slight pause is that some event on EB taking place in the milliseconds that the transfer of baseList.pl takes, in which the TCP connection is closed by the OS for some reason. Granted, it's a very small concern.

The other thing that mitigates any of these concerns is that the mlab-ns update process happens every minute. So, if for some reason the output of baseList.pl were truncated, then notwithstanding some persistent issue on EB, mlab-ns would correct itself less than 60 seconds later.

It seems we have 4 somewhat reasonable choices:

* Scrap this PR, leave mlab-ns as is, and just be as careful as we can that a site/node/sliver never gets removed from Nagios or sites.py without the service/node first being lame-ducked or otherwise flagged as down.

* Accept the small (infinitesimal?) risk that baseList.pl output could conceivable be truncated, and merge the PR, which adds one more small safeguard against human error in the lame-duck process.

* Modify the final line of output of baseList.pl to have a specific sequence of whitespace chars (e.g., \n\t\n\t), and then modify the mlab-ns code to be sure that those are the final 4 chars of the output it gets from baseList.pl before it proceeds.

* Modify mlab-ns code to do some minimal record count of what is returned from baseList.pl, or to look for the presence of strings we know will be near the end (e.g., yyz, wlg, etc.).

Thoughts, @stephen-soltesz @pboothe ?

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/mlab-ns/102#-:-Kr2RR6-H4FBH3B7S7Mq:bmo0tnl)*

stephen-soltesz:
If the nagios config is trustworthy then I'm comfortable trusting baseList.pl. I'm not worried about truncated results during a transfer.

If we don't trust the nagios config, then the final line change could still fail if the nagios config were truncated accidentally (an empty list could still have a valid final line).

The last-site sanity check seems pretty sensible today. I had the impression that it was a hack yesterday -- and maybe it is a little -- but it checks content in two ways: number and value (assuming sites are ordered alphabetically by baselist).

So, my thought is that either we trust the nagios config and this change and baselist are fine as-is. Or, we don't and adding a last-site check to mlab-ns gives a little more safely.

pboothe:
As the least professionally-cautious person, and one who will likely not be developing the solution, I vote for whatever you two work out.

nkinkade:
I'm closing this PR because it seems that the best option may be to just leave the current functionality of not touching datastore records when a corresponding entry does not exist in the Nagios output. The issue this PR was attempting to address can be worked around by simply being sure to lame-duck any node before removing it from Nagios/sites.py.

PR #128, opened 2018-03-30
Allows service status checks to use data from Prometheus, or Nagios.

nkinkade:
We desperately want to stop having production dependencies on Nagios. We are already using Prometheus for a lot of things. One big piece of production code still relying on Nagios is mlab-ns. The PR implements the ability for mlab-ns to update experiment status using Prometheus or Nagios.

A non-goal of this change was to make mlab-ns code better. Instead, the goal was to touch as little of the code as possible, and to change as little of the current code flow as possible. The code may not be great, and flow may not be efficient or ideal, but we _mostly_ trust it to do the right thing. I wanted to avoid making any changes that might introduce subtle bugs that went undetected. We plan on replacing the current mlab-ns codebase with a rewrite at some point in the future anyway.

The implementation here introduces some amount of code redundancy between Nagios and Prometheus. Again, this was intentional. I wanted to touch the Nagios parts of the code as little as possible. So, for example, instead of having a single function that handles either Nagios or Prometheus, I duplicated the function, leaving the Nagios one untouched (to the extent possible), creating a new one for Prometheus.

This code is currently running in as the 'prometheus' service in mlab-nstesting:

https://prometheus-dot-mlab-nstesting.appspot.com/admin/map/ipv4/all

You can verify that the status data came from Prometheus by look at the `tool_extra` column in the SliverTool datastore kind, which should indicate that it was touched by the Prometheus parts of the code.

There are a few sites that are still not updating properly from Prom for some reason, and I still need to investigate why: acc02, lca01, los01, mpm01, tnr01.

Of particular interest in the PR should be the Prometheus status queries in util/prometheus_status.py. I believe them to be correct and accurate, but this should be verified by someone else.

**NOTE**: I have temporarily removed the `check_status` cron job, which runs against the default service, and uses Nagios. I didn't want it overwriting the Prometheus updates to the datastore every single minute.  One can manually trigger an update with this code by fetching this URL:

https://prometheus-dot-mlab-nstesting.appspot.com/cron/check_status

You can toggle whether mlab-ns fetches it's status from Prometheus or Nagios by modifying the `status_source` field in the Tool table:

https://console.cloud.google.com/datastore/entities/query?project=mlab-nstesting&ns=&kind=Tool

Currently, I have set all tools that we care about to get their status from Prometheus to put this through its paces, but if/when we deploy this we'll want to start with a single service, probably neubot, leaving the rest with Nagios during the trial.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/128)

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ To deploy mlab-ns in a fresh GCP project, it is necessary to first deploy the co
 
 Note: These instructions require billing to be enabled on your GCP project, as the data population process will exhaust a free-tiered project's daily AppEngine quota.
 
-Note: These instructions require you to have `nagios.csv`, which is not under source control as it contains secret credentials, but is available to authorized users here: https://goo.gl/tfEg1v.
+Note: These instructions require you to have files named `nagios.csv` and `prometheus.csv`, which are not under source control as they contains secret credentials, but is available to authorized users here: https://goo.gl/tfEg1v. Manually create those two files and paste the file content found at that URL into the files before running the `appcfg.py` commands below.
 
 ```
 # Replace URL with other project's URL if not populating mlab-nstesting
@@ -73,6 +73,11 @@ appcfg.py --url ${GAE_URL}/_ah/remote_api upload_data \
   --config_file=server/bulkloader.yaml \
   --filename=server/mlabns/conf/nagios.csv \
   --kind=Nagios
+
+appcfg.py --url ${GAE_URL}/_ah/remote_api upload_data \
+  --config_file=server/bulkloader.yaml \
+  --filename=server/mlabns/conf/prometheus.csv \
+  --kind=Prometheus
 ```
 
 Note: If you see repeated errors including `Refreshing due to a 401 (attempt
--- a/server/bulkloader.yaml
+++ b/server/bulkloader.yaml
@@ -228,6 +228,9 @@ transformers:
     - property: http_port
       external_name: http_port
 
+    - property: source
+      external_name: source
+
     - property: show_tool_extra
       external_name: show_tool_extra
       import_transform: transform.regexp_bool('true', re.IGNORECASE)
--- a/server/bulkloader.yaml
+++ b/server/bulkloader.yaml
@@ -95,6 +95,26 @@ transformers:
       external_name: url
 
 
+- kind: Prometheus
+  connector: csv
+  connector_options:
+      encoding: utf-8
+  property_map:
+    - property: __key__
+      external_name: index
+      import_transform: "lambda x: 'default_prometheus'"
+      export_transform: transform.key_id_or_name_as_string
+
+    - property: username
+      external_name: username
+
+    - property: password
+      external_name: password
+
+    - property: url
+      external_name: url
+
+
 - kind: Site
   connector: csv
   connector_options:
--- a/server/bulkloader.yaml
+++ b/server/bulkloader.yaml
@@ -102,7 +102,7 @@ transformers:
   property_map:
     - property: __key__
       external_name: index
-      import_transform: "lambda x: 'default_prometheus'"
+      import_transform: "lambda x: 'prometheus_default'"
       export_transform: transform.key_id_or_name_as_string
 
     - property: username
--- a/server/bulkloader.yaml
+++ b/server/bulkloader.yaml
@@ -228,8 +228,8 @@ transformers:
     - property: http_port
       external_name: http_port
 
-    - property: source
-      external_name: source
+    - property: status_source
+      external_name: status_source
 
     - property: show_tool_extra
       external_name: show_tool_extra
--- a/server/mlabns/conf/tools.csv
+++ b/server/mlabns/conf/tools.csv
@@ -1,4 +1,4 @@
-tool_id,slice_id,http_port,show_tool_extra,source
+tool_id,slice_id,http_port,show_tool_extra,status_source
 ndt,iupui_ndt,7123,False,nagios
 ndt_ssl,iupui_ndt,,False,nagios
 mobiperf,michigan_1,,False,nagios
--- a/server/mlabns/conf/tools.csv
+++ b/server/mlabns/conf/tools.csv
@@ -1,7 +1,7 @@
-tool_id,slice_id,http_port,show_tool_extra
-ndt,iupui_ndt,7123,False
-ndt_ssl,iupui_ndt,,False
-mobiperf,michigan_1,,False
-npad,iupui_npad,8000,False
-ooni,mlab_ooni,,True
-neubot,mlab_neubot,8080,False
+tool_id,slice_id,http_port,show_tool_extra,source
+ndt,iupui_ndt,7123,False,nagios
+ndt_ssl,iupui_ndt,,False,nagios
+mobiperf,michigan_1,,False,nagios
+npad,iupui_npad,8000,False,nagios
+ooni,mlab_ooni,,True,nagios
+neubot,mlab_neubot,8080,False,prometheus
--- a/server/mlabns/db/model.py
+++ b/server/mlabns/db/model.py
@@ -203,3 +203,21 @@ def get_all_tool_ids():
     for tool in Tool.all().run(batch_size=constants.GQL_BATCH_SIZE):
         tool_ids.append(tool.tool_id)
     return tool_ids
+
+
+def get_status_source_deps(source):
+    """Gets all Tools that rely on the given status source.
+
+    Args:
+        source: str, status source, either 'nagios or 'prometheus'.
+
+    Returns:
+        list, Tools that depend on `source`.
+    """
+    tools = []
+    tools_gql = Tool.gql("WHERE status_source = :source", source=source)
+    for tool in tools_gql.run():
+        tools.append(tool)
+    if not tools:
+        logging.info('No tools get their status from %s.', source)
+    return tools
--- a/server/mlabns/db/model.py
+++ b/server/mlabns/db/model.py
@@ -124,9 +124,17 @@ class Tool(db.Model):
     # redirected to: http://fqdn[ipv4|ipv6]:http_port
     http_port = db.StringProperty()
     show_tool_extra = db.BooleanProperty()
+    status_source = db.StringProperty()
 
 
 class Nagios(db.Model):
+    key_id = db.StringProperty()
+    username = db.StringProperty()
+    password = db.StringProperty()
+    url = db.StringProperty()
+
+
+class Prometheus(db.Model):
     key_id = db.StringProperty()
     username = db.StringProperty()
     password = db.StringProperty()
--- a/server/mlabns/db/prometheus_config_wrapper.py
+++ b/server/mlabns/db/prometheus_config_wrapper.py
@@ -1,0 +1,23 @@
+import logging
+
+from google.appengine.api import memcache
+
+from mlabns.db import model
+from mlabns.util import constants
+
+
+def get_prometheus_config():
+    """Retrieves Prometheus config info. First checks memcache, then datastore.
+
+    Returns:
+        Prometheus model instance
+    """
+    prometheus = memcache.get(constants.DEFAULT_PROMETHEUS_ENTRY)
+    if not prometheus:
+        prometheus = model.Prometheus.get_by_key_name(constants.DEFAULT_PROMETHEUS_ENTRY)
+        if prometheus:
+            memcache.set(constants.DEFAULT_PROMETHEUS_ENTRY, prometheus)
+        else:
+            logging.error('Datastore does not have the Prometheus credentials.')
+
+    return prometheus
--- a/server/mlabns/db/prometheus_config_wrapper.py
+++ b/server/mlabns/db/prometheus_config_wrapper.py
@@ -14,7 +14,8 @@ def get_prometheus_config():
     """
     prometheus = memcache.get(constants.DEFAULT_PROMETHEUS_ENTRY)
     if not prometheus:
-        prometheus = model.Prometheus.get_by_key_name(constants.DEFAULT_PROMETHEUS_ENTRY)
+        prometheus = model.Prometheus.get_by_key_name(
+            constants.DEFAULT_PROMETHEUS_ENTRY)
         if prometheus:
             memcache.set(constants.DEFAULT_PROMETHEUS_ENTRY, prometheus)
         else:
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -380,15 +380,15 @@ class StatusUpdateHandler(webapp.RequestHandler):
             tool = model.get_tool_from_tool_id(tool_id)
             for address_family in ['', '_ipv6']:
                 if tool.status_source == 'prometheus':
-                    logging.info('Status source for %s is: prometheus' %
-                                 tool_id)
+                    logging.info('Status source for %s%s is: prometheus' %
+                                 tool_id, address_family)
                     slice_info = prometheus_status.get_slice_info(
                         prometheus_config.url, tool_id, address_family)
-                    logging.warning('PROM URL: %s' % slice_info.slice_url)
                     slice_status = prometheus_status.get_slice_status(
                         slice_info.slice_url, prometheus_opener, tool.slice_id)
                 elif tool.status_source == 'nagios':
-                    logging.info('Status source for %s is: nagios' % tool_id)
+                    logging.info('Status source for %s%s is: nagios' % tool_id,
+                            address_family)
                     slice_info = nagios_status.get_slice_info(
                         nagios_config.url, tool_id, address_family)
                     slice_status = nagios_status.get_slice_status(
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -354,7 +354,6 @@ class StatusUpdateHandler(webapp.RequestHandler):
             logging.error('Datastore does not have the Prometheus configs.')
             return util.send_not_found(self)
         prometheus_opener = prometheus_status.authenticate_prometheus(prometheus_config)
-        logging.info('FIRST prometheus_config.url: %s' % prometheus_config.url)
 
         # Get Nagios configs, and authenticate.
         nagios_config = nagios_config_wrapper.get_nagios_config()
@@ -362,9 +361,6 @@ class StatusUpdateHandler(webapp.RequestHandler):
             logging.error('Datastore does not have the Nagios configs.')
             return util.send_not_found(self)
         nagios_opener = nagios_status.authenticate_nagios(nagios_config)
-        logging.info('nagios_config.url: %s' % nagios_config.url)
-
-        logging.info('SECOND prometheus_config.url: %s' % prometheus_config.url)
 
         for tool_id in model.get_all_tool_ids():
             tool = model.get_tool_from_tool_id(tool_id)
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -350,32 +350,33 @@ class StatusUpdateHandler(webapp.RequestHandler):
         """
         for tool_id in model.get_all_tool_ids():
             tool = model.get_tool_from_tool_id(tool_id)
-            if tool.status_source == 'prometheus':
-                logging.info('Status source for %s is: prometheus' % tool_id)
-                prometheus = prometheus_config_wrapper.get_prometheus_config()
-                if prometheus is None:
-                    logging.error('Datastore does not have the Prometheus credentials.')
-                    return util.send_not_found(self)
-                prometheus_status.authenticate_prometheus(prometheus)
-                slice_info = prometheus_status.get_slice_info(prometheus.url)
-                slice_status = prometheus_status.get_slice_status(
-                        slice_info.slice_url, tool.slice_id)
-            elif tool.status_source == 'nagios':
-                logging.info('Status source for %s is: nagios' % tool_id)
-                nagios = nagios_config_wrapper.get_nagios_config()
-                if nagios is None:
-                    logging.error('Datastore does not have the Nagios credentials.')
-                    return util.send_not_found(self)
-                nagios_status.authenticate_nagios(nagios)
-                slice_info = nagios_status.get_slice_info(nagios.url)
-                slice_status = nagios_status.get_slice_status(slice_info.slice_url)
-            else:
-                logging.error('Unknown tool status_source: %s.', status_source)
-                continue
+            for address_family in ['', '_ipv6']:
+                if tool.status_source == 'prometheus':
+                    logging.info('Status source for %s is: prometheus' % tool_id)
+                    prometheus = prometheus_config_wrapper.get_prometheus_config()
+                    if prometheus is None:
+                        logging.error('Datastore does not have the Prometheus credentials.')
+                        return util.send_not_found(self)
+                    prometheus_status.authenticate_prometheus(prometheus)
+                    slice_info = prometheus_status.get_slice_info(prometheus.url, address_family)
+                    slice_status = prometheus_status.get_slice_status(
+                            slice_info.slice_url, tool.slice_id)
+                elif tool.status_source == 'nagios':
+                    logging.info('Status source for %s is: nagios' % tool_id)
+                    nagios = nagios_config_wrapper.get_nagios_config()
+                    if nagios is None:
+                        logging.error('Datastore does not have the Nagios credentials.')
+                        return util.send_not_found(self)
+                    nagios_status.authenticate_nagios(nagios)
+                    slice_info = nagios_status.get_slice_info(nagios.url, address_family)
+                    slice_status = nagios_status.get_slice_status(slice_info.slice_url)
+                else:
+                    logging.error('Unknown tool status_source: %s.', status_source)
+                    continue
 
-            if slice_status:
-                self.update_sliver_tools_status(
-                    slice_status, slice_info.tool_id, slice_info.address_family)
+                if slice_status:
+                    self.update_sliver_tools_status(
+                        slice_status, slice_info.tool_id, slice_info.address_family)
 
         return util.send_success(self)
 
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -385,7 +385,7 @@ class StatusUpdateHandler(webapp.RequestHandler):
                     slice_info = prometheus_status.get_slice_info(
                         prometheus_config.url, tool_id, address_family)
                     slice_status = prometheus_status.get_slice_status(
-                        slice_info.slice_url, prometheus_opener, tool.slice_id)
+                        slice_info.slice_url, prometheus_opener)
                 elif tool.status_source == 'nagios':
                     logging.info('Status source for %s%s is: nagios' % tool_id,
                             address_family)
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -358,7 +358,7 @@ class StatusUpdateHandler(webapp.RequestHandler):
                         logging.error('Datastore does not have the Prometheus credentials.')
                         return util.send_not_found(self)
                     prometheus_status.authenticate_prometheus(prometheus)
-                    slice_info = prometheus_status.get_slice_info(prometheus.url, address_family)
+                    slice_info = prometheus_status.get_slice_info(prometheus.url, tool_id, address_family)
                     slice_status = prometheus_status.get_slice_status(
                             slice_info.slice_url, tool.slice_id)
                 elif tool.status_source == 'nagios':
@@ -368,7 +368,7 @@ class StatusUpdateHandler(webapp.RequestHandler):
                         logging.error('Datastore does not have the Nagios credentials.')
                         return util.send_not_found(self)
                     nagios_status.authenticate_nagios(nagios)
-                    slice_info = nagios_status.get_slice_info(nagios.url, address_family)
+                    slice_info = nagios_status.get_slice_info(nagios.url, tool_id, address_family)
                     slice_status = nagios_status.get_slice_status(slice_info.slice_url)
                 else:
                     logging.error('Unknown tool status_source: %s.', status_source)
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -368,12 +368,10 @@ class StatusUpdateHandler(webapp.RequestHandler):
                 if tool.status_source == 'prometheus':
                     logging.info('Status source for %s is: prometheus' % tool_id)
                     slice_info = prometheus_status.get_slice_info(prometheus_config.url, tool_id, address_family)
-                    logging.info('Status URL for %s is: %s' % (tool_id, slice_info.slice_url))
                     slice_status = prometheus_status.get_slice_status(slice_info.slice_url, prometheus_opener, tool.slice_id)
                 elif tool.status_source == 'nagios':
                     logging.info('Status source for %s is: nagios' % tool_id)
                     slice_info = nagios_status.get_slice_info(nagios_config.url, tool_id, address_family)
-                    logging.info('Status URL for %s is: %s' % (tool_id, slice_info.slice_url))
                     slice_status = nagios_status.get_slice_status(slice_info.slice_url, nagios_opener)
                 else:
                     logging.error('Unknown tool status_source: %s.', status_source)
@@ -402,7 +400,7 @@ class StatusUpdateHandler(webapp.RequestHandler):
         for sliver_tool in sliver_tools:
 
             if sliver_tool.fqdn not in slice_status:
-                logging.info('Prometheus does not know sliver %s.',
+                logging.info('Monitoring does not know sliver %s.',
                              sliver_tool.fqdn)
                 continue
 
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -388,6 +388,8 @@ class StatusUpdateHandler(webapp.RequestHandler):
                     if prometheus_config:
                         slice_info = prometheus_status.get_slice_info(
                             prometheus_config.url, tool_id, address_family)
+                        if not slice_info:
+                            continue
                         slice_status = prometheus_status.get_slice_status(
                             slice_info.slice_url, prometheus_opener)
                     else:
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -345,8 +345,8 @@ class StatusUpdateHandler(webapp.RequestHandler):
         """Triggers the update handler.
 
         Updates sliver status with information from either Nagios or Prometheus.
-        The URLs containing the information are stored in the datastore along
-        with the credentials necessary to access the data.
+        The base URLs for accessing status information are stored in the
+        datastore along with the credentials necessary to access the data.
         """
         # Get Prometheus configs, and authenticate.
         prometheus_config = prometheus_config_wrapper.get_prometheus_config()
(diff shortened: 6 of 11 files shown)


nkinkade:
**TODO**: Add new unittest for prometheus_status.py, and other Prom-related things.

stephen-soltesz:
Reviewed 6 of 9 files at r1.
Review status: 6 of 9 files reviewed at latest revision, all discussions resolved.

---

*[server/bulkloader.yaml, line 98 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L96irI32V5K9NFhA5in:-L96irI32V5K9NFhA5io:bxx22e4) ([raw file](https://github.com/m-lab/mlab-ns/blob/cebacc04ef201587d593bcb04b4f5589e0e072b6/server/bulkloader.yaml#L98)):*
> ```YAML
> 
> 
> - kind: Prometheus
> ```

This is the config for the prometheus credentials that's loaded once?

Can we add an example file? Or is it already documented somewhere else for the nagios format? We should plan to update those docs to refer to prometheus by default.

---

*[server/mlabns/handlers/update.py, line 353 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L96YZB38sYy11nigQtU:-L96YZB38sYy11nigQtV:bhve4l4) ([raw file](https://github.com/m-lab/mlab-ns/blob/cebacc04ef201587d593bcb04b4f5589e0e072b6/server/mlabns/handlers/update.py#L353)):*
> ```Python
>         # Get Prometheus configs, and authenticate.
>         prometheus_config = prometheus_config_wrapper.get_prometheus_config()
>         if prometheus_config is None:
> ```

Would it be expensive to make this check (and the one for nagios) additionally conditional on whether at least one experiment has a 'status_source'  for this service?

My thinking being that we don't want to return an error unless at least one tool depends on this source. Also, I'm wondering if we should still try both without error if one is missing so that some tools are updated.

This may be optional. Raising for consideration.

---

*[server/mlabns/util/prometheus_status.py, line 68 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L96b-8_9Z6lrSQBYwtM:-L96b-8_9Z6lrSQBYwtN:b1ceb61) ([raw file](https://github.com/m-lab/mlab-ns/blob/cebacc04ef201587d593bcb04b4f5589e0e072b6/server/mlabns/util/prometheus_status.py#L68)):*
> ```Python
>         A urllib2 OpenerDirector object.
>     """
>     password_manager = urllib2.HTTPPasswordMgrWithDefaultRealm()
> ```

Does this work?

I thought basic auth was not the same as digest auth. But, I'm a little fuzzy about it.

Is the prometheus.url using the new TLS target address?

If not, then I think we want to use the TLS target. And, that may explain why the digest with empty user/passwd may be working today.

---

*[server/mlabns/util/prometheus_status.py, line 92 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L96gD1WApdJ6r24ZlQ0:-L96gD1WApdJ6r24ZlQ1:b-5oz961) ([raw file](https://github.com/m-lab/mlab-ns/blob/cebacc04ef201587d593bcb04b4f5589e0e072b6/server/mlabns/util/prometheus_status.py#L92)):*
> ```Python
>     # Joins the experiment name with the machine name to form the FQDN fo the
>     # experiment.
>     sliver_fqdn = experiment + '.' + status['metric']['machine']
> ```

Just a thought. If the prometheus queries grouped by `(experiment, machine)` then both parts of the sliver fqdn would be returned. This would avoid the `experiment` reconstruction above.

---

*[server/mlabns/util/prometheus_status.py, line 93 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L96i1Lh26ELC2fovCs-:-L96i1Lh26ELC2fovCs0:blbowte) ([raw file](https://github.com/m-lab/mlab-ns/blob/cebacc04ef201587d593bcb04b4f5589e0e072b6/server/mlabns/util/prometheus_status.py#L93)):*
> ```Python
>     # experiment.
>     sliver_fqdn = experiment + '.' + status['metric']['machine']
>     state = status['value'][1]
> ```

Please add a hint that the 'value' is a (timestamp, state) tuple, and we only want the  state value.

---

*[server/mlabns/util/prometheus_status.py, line 116 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L96bkHM83nwR93HUDoy:-L96bkHM83nwR93HUDoz:blz8i5b) ([raw file](https://github.com/m-lab/mlab-ns/blob/cebacc04ef201587d593bcb04b4f5589e0e072b6/server/mlabns/util/prometheus_status.py#L116)):*
> ```Python
>     # return the status for the tool.
>     queries = {
>         'ndt': ('min by (machine) ( '
> ```

I expect we will need to update these queries, so making that a little easier will help us in the future.

I suggest using something like [textwrap dedent][1] to preserve a single quoted string with natural indent here. That will make formatting of the query as well as updates much easier.

I'll carefully review each query when I can easily copy/paste the queries into prometheus.

[1]: https://github.com/m-lab/scraper/blob/24d5cd52592be50046950e50d67071464e4ef02a/scraper_test.py#L125

---

*[server/mlabns/util/prometheus_status.py, line 188 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L96ccHh0sStm0jYItRM:-L96ccHh0sStm0jYItRN:b8imf8g) ([raw file](https://github.com/m-lab/mlab-ns/blob/cebacc04ef201587d593bcb04b4f5589e0e072b6/server/mlabns/util/prometheus_status.py#L188)):*
> ```Python
>         return None
> 
>     statuses = json.loads(raw_data)
> ```

I suggest also handling `ValueError` here in case the open/read above successfully returns an invalid json object. this could happen for example if the loadbalancer redirects to a "hello world" http target.

---

*[server/mlabns/util/prometheus_status.py, line 191 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L96d-F0EWGHowTPnyXU:-L96d-F0EWGHowTPnyXV:bxheyl) ([raw file](https://github.com/m-lab/mlab-ns/blob/cebacc04ef201587d593bcb04b4f5589e0e072b6/server/mlabns/util/prometheus_status.py#L191)):*
> ```Python
> 
>     if statuses['status'] == 'error':
>         logging.info('Prometheus returned error "%s" for URL %s.' %
> ```

`logging.error`?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128)*

nkinkade:
@stephen-soltesz: I've made a number of commits to address some of the things you brought up, and I've replied to your comments. PTAL?

---

Review status: 5 of 10 files reviewed at latest revision, 8 unresolved discussions.

---

*[server/bulkloader.yaml, line 98 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L96irI32V5K9NFhA5in:-L9C4GB66TR8XQTVN1ks:b-pybegk) ([raw file](https://github.com/m-lab/mlab-ns/blob/cebacc04ef201587d593bcb04b4f5589e0e072b6/server/bulkloader.yaml#L98)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

This is the config for the prometheus credentials that's loaded once?

Can we add an example file? Or is it already documented somewhere else for the nagios format? We should plan to update those docs to refer to prometheus by default.
</blockquote></details>

It is loaded once when mlab-ns is first instantiated. There are instructions in the README for handling the Nagios credentials and file. I've updated the README, along with the credentials file in Google Drive to include instructions for Prometheus.

---

*[server/mlabns/handlers/update.py, line 353 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L96YZB38sYy11nigQtU:-L9C1nzO264V45K6q4q1:be6qg8o) ([raw file](https://github.com/m-lab/mlab-ns/blob/cebacc04ef201587d593bcb04b4f5589e0e072b6/server/mlabns/handlers/update.py#L353)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

Would it be expensive to make this check (and the one for nagios) additionally conditional on whether at least one experiment has a 'status_source'  for this service?

My thinking being that we don't want to return an error unless at least one tool depends on this source. Also, I'm wondering if we should still try both without error if one is missing so that some tools are updated.

This may be optional. Raising for consideration.
</blockquote></details>

This makes sense. I've implemented something that I think might serve the purpose. PTAL?

---

*[server/mlabns/util/prometheus_status.py, line 68 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L96b-8_9Z6lrSQBYwtM:-L97XuPCDcXVq9f3XU0N:b-3fmz0r) ([raw file](https://github.com/m-lab/mlab-ns/blob/cebacc04ef201587d593bcb04b4f5589e0e072b6/server/mlabns/util/prometheus_status.py#L68)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

Does this work?

I thought basic auth was not the same as digest auth. But, I'm a little fuzzy about it.

Is the prometheus.url using the new TLS target address?

If not, then I think we want to use the TLS target. And, that may explain why the digest with empty user/passwd may be working today.
</blockquote></details>

I've got the current credentials in the datastore, but was using the old status.* URL, since until just today the new authentication and URLs were not in place in production. Since you deployed these changes to production today I can start using the new https URL along with the credentials.

Regarding Basic Auth vs Digest, I think you are right that there is definitely a difference that makes them incompatible. I've changed the code here for Prometheus to use the HTTPBasicAuthHandler() instead. Thanks for catching this.

---

*[server/mlabns/util/prometheus_status.py, line 92 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L96gD1WApdJ6r24ZlQ0:-L97qg-p0kMssZMMxoMq:b77l5lx) ([raw file](https://github.com/m-lab/mlab-ns/blob/cebacc04ef201587d593bcb04b4f5589e0e072b6/server/mlabns/util/prometheus_status.py#L92)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

Just a thought. If the prometheus queries grouped by `(experiment, machine)` then both parts of the sliver fqdn would be returned. This would avoid the `experiment` reconstruction above.
</blockquote></details>

This is a great idea and would simplify the code a good bit in multiple places (because I currently have to pass around `tool.slice_id` for this one purpose). However, I don't think it will work 100% because of the `lame_duck_host{}` part of the NDT queries. There is no `experiment` label in those metrics (but there is a `machine` label), so grouping by `(experiment, machine)` would eliminate any `lame_duck_node{}` results, meaning we couldn't take lame-ducked notes into account.

One alternative, which could make this work, would be to modify how do lame-duck signaling. We could add experiment-level lame-duck metrics. This might be desirable irrespective of this particular use case. What do you think?

---

*[server/mlabns/util/prometheus_status.py, line 93 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L96i1Lh26ELC2fovCs-:-L97dNdi7SPItsN-aGaw:bwko1cn) ([raw file](https://github.com/m-lab/mlab-ns/blob/cebacc04ef201587d593bcb04b4f5589e0e072b6/server/mlabns/util/prometheus_status.py#L93)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

Please add a hint that the 'value' is a (timestamp, state) tuple, and we only want the  state value.
</blockquote></details>

Note added.

---

*[server/mlabns/util/prometheus_status.py, line 116 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L96bkHM83nwR93HUDoy:-L97jM7jDnCAN3FTakcx:b-896fix) ([raw file](https://github.com/m-lab/mlab-ns/blob/cebacc04ef201587d593bcb04b4f5589e0e072b6/server/mlabns/util/prometheus_status.py#L116)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

I expect we will need to update these queries, so making that a little easier will help us in the future.

I suggest using something like [textwrap dedent][1] to preserve a single quoted string with natural indent here. That will make formatting of the query as well as updates much easier.

I'll carefully review each query when I can easily copy/paste the queries into prometheus.

[1]: https://github.com/m-lab/scraper/blob/24d5cd52592be50046950e50d67071464e4ef02a/scraper_test.py#L125
</blockquote></details>

Done.

---

*[server/mlabns/util/prometheus_status.py, line 188 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L96ccHh0sStm0jYItRM:-L97kxW67Xa1wcB4iqPT:bm3m4vx) ([raw file](https://github.com/m-lab/mlab-ns/blob/cebacc04ef201587d593bcb04b4f5589e0e072b6/server/mlabns/util/prometheus_status.py#L188)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

I suggest also handling `ValueError` here in case the open/read above successfully returns an invalid json object. this could happen for example if the loadbalancer redirects to a "hello world" http target.
</blockquote></details>

This makes senses. Done.

---

*[server/mlabns/util/prometheus_status.py, line 191 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L96d-F0EWGHowTPnyXU:-L97kzOm2pfNoG_AT2c0:bt2vr1b) ([raw file](https://github.com/m-lab/mlab-ns/blob/cebacc04ef201587d593bcb04b4f5589e0e072b6/server/mlabns/util/prometheus_status.py#L191)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

`logging.error`?
</blockquote></details>

Fixed.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-:-L9C1uhq8VUhxovi0_Z4:b-hlosw6)*

stephen-soltesz:
The lame_duck_experiment changes are looking good. Added a few more comments.

---

Reviewed 1 of 9 files at r1, 2 of 4 files at r2.
Review status: 8 of 10 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[server/mlabns/handlers/update.py, line 382 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9CQjf45_SmkQG3O_X5:-L9CQjf45_SmkQG3O_X6:b23jsof) ([raw file](https://github.com/m-lab/mlab-ns/blob/718d820f475cc7ab0cf6d35ca83f3f681df46f6b/server/mlabns/handlers/update.py#L382)):*
> ```Python
>             tool = model.get_tool_from_tool_id(tool_id)
>             for address_family in ['', '_ipv6']:
>                 if tool.status_source == 'prometheus':
> ```

It looks like prometheus_opener will be undefined when prometheus_config is None. What's the best way to handle this?

---

*[server/mlabns/handlers/update.py, line 383 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9CPflc8fvtzO3cCB1g:-L9CPflc8fvtzO3cCB1h:bq6fzno) ([raw file](https://github.com/m-lab/mlab-ns/blob/718d820f475cc7ab0cf6d35ca83f3f681df46f6b/server/mlabns/handlers/update.py#L383)):*
> ```Python
>             for address_family in ['', '_ipv6']:
>                 if tool.status_source == 'prometheus':
>                     logging.info('Status source for %s%s is: prometheus' %
> ```

For logging with format parameters, prefer calling as:
```
logging.info("message with format %s", value)
```

This takes some getting used to. Since logging level is conditional, in some cases the formatting may be skipped. And, by passing the values as parameters, we leave the interpolation of the format string to the logging function, which should always do the right thing.

---

*[server/mlabns/util/prometheus_status.py, line 92 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L96gD1WApdJ6r24ZlQ0:-L9C6kN70ZIQElzkH0CY:b-s248a4) ([raw file](https://github.com/m-lab/mlab-ns/blob/cebacc04ef201587d593bcb04b4f5589e0e072b6/server/mlabns/util/prometheus_status.py#L92)):*

This is a great idea and would simplify the code a good bit in multiple places (because I currently have to pass around `tool.slice_id` for this one purpose). However, I don't think it will work 100% because of the `lame_duck_host{}` part of the NDT queries. There is no `experiment` label in those metrics (but there is a `machine` label), so grouping by `(experiment, machine)` would eliminate any `lame_duck_node{}` results, meaning we couldn't take lame-ducked notes into account.

One alternative, which could make this work, would be to modify how do lame-duck signaling. We could add experiment-level lame-duck metrics. This might be desirable irrespective of this particular use case. What do you think?
</blockquote></details>

We discussed in person, and you're adding a per-experiment lame_duck metric that will always be present.

---

*[server/mlabns/util/prometheus_status.py, line 126 at r2](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9CHiexEGRp9D4v2f8G:-L9CHiexEGRp9D4v2f8H:bla4rgm) ([raw file](https://github.com/m-lab/mlab-ns/blob/33922c785483aacbcdc6b4edd28c5c240a73c79e/server/mlabns/util/prometheus_status.py#L126)):*
> ```Python
>               (vdlimit_used{experiment="ndt.iupui"} /
>                   vdlimit_total{experiment="ndt.iupui"}) < bool 0.95 OR
>               lame_duck_node{} != bool 1)"""),
> ```

nit: please place the end quote on it's own line.  This again makes it a little easier to copy the query out.

---

*[server/mlabns/util/prometheus_status.py, line 198 at r2](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9CO-tmApJWBV9Pvjnq:-L9CO-tmApJWBV9Pvjnr:byjz7bd) ([raw file](https://github.com/m-lab/mlab-ns/blob/33922c785483aacbcdc6b4edd28c5c240a73c79e/server/mlabns/util/prometheus_status.py#L198)):*
> ```Python
>         statuses = json.loads(raw_data)
>     except ValueError:
>         logging.error('Unable to parse JSON from: %s' % url)
> ```

And, return?

---

*[server/mlabns/util/prometheus_status.py, line 122 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9CR4lF2XiLESgVCwO8:-L9CR4lF2XiLESgVCwO9:b-255zxy) ([raw file](https://github.com/m-lab/mlab-ns/blob/718d820f475cc7ab0cf6d35ca83f3f681df46f6b/server/mlabns/util/prometheus_status.py#L122)):*
> ```Python
>               (vdlimit_used{experiment="ndt.iupui"} /
>                   vdlimit_total{experiment="ndt.iupui"}) < bool 0.95 OR
>               lame_duck_experiment{} != bool 1)"""),
> ```

Since we have lame_duck_experiment metrics for every experiment now, also select for `experiment="ndt.iupui"`.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-:-L9CS6z99OH9Mmg2UB2g:b12f8km)*

nkinkade:
I've made some additional commits to address the issues you raised. PTAL?

---

Review status: 8 of 10 files reviewed at latest revision, 7 unresolved discussions.

---

*[server/mlabns/handlers/update.py, line 382 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9CQjf45_SmkQG3O_X5:-L9CcxxHAJhTh_0_buWp:b-juf13i) ([raw file](https://github.com/m-lab/mlab-ns/blob/718d820f475cc7ab0cf6d35ca83f3f681df46f6b/server/mlabns/handlers/update.py#L382)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

It looks like prometheus_opener will be undefined when prometheus_config is None. What's the best way to handle this?
</blockquote></details>

I think I've addressed this with commit @f7314da.  The whole thing is starting to feeling a bit crufty, but it sort of was before, too.

---

*[server/mlabns/handlers/update.py, line 383 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9CPflc8fvtzO3cCB1g:-L9CY9xeClNx4Nqg28Ic:b-1wp4j8) ([raw file](https://github.com/m-lab/mlab-ns/blob/718d820f475cc7ab0cf6d35ca83f3f681df46f6b/server/mlabns/handlers/update.py#L383)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

For logging with format parameters, prefer calling as:
```
logging.info("message with format %s", value)
```

This takes some getting used to. Since logging level is conditional, in some cases the formatting may be skipped. And, by passing the values as parameters, we leave the interpolation of the format string to the logging function, which should always do the right thing.
</blockquote></details>

Fixed, in both this file and prometheus_status.py.

---

*[server/mlabns/util/prometheus_status.py, line 92 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L96gD1WApdJ6r24ZlQ0:-L9CXAM8EFP5OpPMvw1Q:b-4818yi) ([raw file](https://github.com/m-lab/mlab-ns/blob/cebacc04ef201587d593bcb04b4f5589e0e072b6/server/mlabns/util/prometheus_status.py#L92)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

We discussed in person, and you're adding a per-experiment lame_duck metric that will always be present.
</blockquote></details>

And now they are! I think this is fixed.

---

*[server/mlabns/util/prometheus_status.py, line 126 at r2](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9CHiexEGRp9D4v2f8G:-L9CWimcAo2uUCIWeKri:bt2vr1b) ([raw file](https://github.com/m-lab/mlab-ns/blob/33922c785483aacbcdc6b4edd28c5c240a73c79e/server/mlabns/util/prometheus_status.py#L126)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

nit: please place the end quote on it's own line.  This again makes it a little easier to copy the query out.
</blockquote></details>

Fixed.

---

*[server/mlabns/util/prometheus_status.py, line 198 at r2](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9CO-tmApJWBV9Pvjnq:-L9CX-ww4VIug9TFYLtL:bt2vr1b) ([raw file](https://github.com/m-lab/mlab-ns/blob/33922c785483aacbcdc6b4edd28c5c240a73c79e/server/mlabns/util/prometheus_status.py#L198)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

And, return?
</blockquote></details>

Fixed.

---

*[server/mlabns/util/prometheus_status.py, line 122 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9CR4lF2XiLESgVCwO8:-L9CWjfJDC8dBSZHV0nr:bt2vr1b) ([raw file](https://github.com/m-lab/mlab-ns/blob/718d820f475cc7ab0cf6d35ca83f3f681df46f6b/server/mlabns/util/prometheus_status.py#L122)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

Since we have lame_duck_experiment metrics for every experiment now, also select for `experiment="ndt.iupui"`.
</blockquote></details>

Fixed.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-:-L9CdOcNApIXnCxfFjja:bjiuapd)*

nkinkade:
@stephen-soltesz: Some unit tests were added. What do you think?

---

Review status: 7 of 11 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-:-L9HOYPX-trUAmrynLsf:bsc6qr8)*

stephen-soltesz:
Great! I have some style suggestions that I think will help.

---

Reviewed 2 of 2 files at r4, 3 of 4 files at r5.
Review status: 10 of 11 files reviewed at latest revision, all discussions resolved.

---

*[server/mlabns/tests/test_prometheus_status.py, line 41 at r5](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9HScnDEwU3PRjPsLEv:-L9HScnDEwU3PRjPsLEw:b-30fb4p) ([raw file](https://github.com/m-lab/mlab-ns/blob/74163e1b49547e2672ea90b2b3528ea6eee07415/server/mlabns/tests/test_prometheus_status.py#L41)):*
> ```Python
> 
>     def setUp(self):
>         self.prometheus_base_url = 'https://prometheus.mlab-oti.measurementlab.net/api/v1/query?query='
> ```

Since there are no actual network connections, I suggest using a phony domain name to prevent future confusion. I had to double check to make sure there wasn't a live query during the test.

---

*[server/mlabns/tests/test_prometheus_status.py, line 45 at r5](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9HRDukC-A1xOjaAzI0:-L9HRDukC-A1xOjaAzI1:b-iewqdi) ([raw file](https://github.com/m-lab/mlab-ns/blob/74163e1b49547e2672ea90b2b3528ea6eee07415/server/mlabns/tests/test_prometheus_status.py#L45)):*
> ```Python
>     def test_get_slice_info_returns_none_with_nonexistent_tool(self):
>         retrieved = prometheus_status.get_slice_info(self.prometheus_base_url,
>                                                      'lol', '')
> ```

Please rename 'lol' to something like 'nonexistent_tool'.

---

*[server/mlabns/tests/test_prometheus_status.py, line 57 at r5](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9HSVvp26goJp0bCj_3:-L9HSVvp26goJp0bCj_4:b-8sewi8) ([raw file](https://github.com/m-lab/mlab-ns/blob/74163e1b49547e2672ea90b2b3528ea6eee07415/server/mlabns/tests/test_prometheus_status.py#L57)):*
> ```Python
>         neubot_url_ipv6 = self.prometheus_base_url + urllib.quote_plus(
>             prometheus_status.QUERIES['neubot_ipv6'])
>         slice_data = {
> ```

Please name `expected_slice_data`.

---

*[server/mlabns/tests/test_prometheus_status.py, line 72 at r5](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9HaF3j9Egbix2UMW4f:-L9HaF3j9Egbix2UMW4g:b-sok6wm) ([raw file](https://github.com/m-lab/mlab-ns/blob/74163e1b49547e2672ea90b2b3528ea6eee07415/server/mlabns/tests/test_prometheus_status.py#L72)):*
> ```Python
>         }
> 
>         for tool_id in ['ndt', 'neubot']:
> ```

This is a sensible approach.

I've grown up with the bias / convention of "no complex logic in unit tests". The goal being to have as much "straight line" execution as possible.

In this case, I would probably make four explicit blocks with literal parameters, like:
```
retrieved = prometheus_status.get_slice_info(self.prometheus_base_url, 'ndt', '')
self.assertEqual(expected_slice_data['ndt']['info'], retrieved)

# repeat with other variations.
```

---

*[server/mlabns/tests/test_prometheus_status.py, line 83 at r5](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9HURzJCpGDCWi6ofYM:-L9HURzJCpGDCWi6ofYN:b-eqv90n) ([raw file](https://github.com/m-lab/mlab-ns/blob/74163e1b49547e2672ea90b2b3528ea6eee07415/server/mlabns/tests/test_prometheus_status.py#L83)):*
> ```Python
> class StatusUpdateHandlerTest(unittest2.TestCase):
> 
>     def setUp(self):
> ```

Please try to eliminate this style of mock setup unless absolutely unavoidable. See comments below for suggested style.

---

*[server/mlabns/tests/test_prometheus_status.py, line 84 at r5](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9HZcvi5N5LwhyF4ls6:-L9HZcvi5N5LwhyF4ls7:bxqznzi) ([raw file](https://github.com/m-lab/mlab-ns/blob/74163e1b49547e2672ea90b2b3528ea6eee07415/server/mlabns/tests/test_prometheus_status.py#L84)):*
> ```Python
> 
>     def setUp(self):
>         self.status_update_handler = update.StatusUpdateHandler()
> ```

Is this parameter needed? I don't see it used.

---

*[server/mlabns/tests/test_prometheus_status.py, line 86 at r5](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9HZfzw0nKyVOg2I59Q:-L9HZfzw0nKyVOg2I59R:bpr5enu) ([raw file](https://github.com/m-lab/mlab-ns/blob/74163e1b49547e2672ea90b2b3528ea6eee07415/server/mlabns/tests/test_prometheus_status.py#L86)):*
> ```Python
>         self.status_update_handler = update.StatusUpdateHandler()
>         self.mock_urlopen_response = mock.Mock()
>         self.opener = urllib2.OpenerDirector()
> ```

Reusing variables like this across functions can be risky in the event that any function modifies it in an unknown way, all other functions that use it may see unexpected sideffects. This can be okay if the variable is clearly read-only.

---

*[server/mlabns/tests/test_prometheus_status.py, line 89 at r5](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9HYuMKBvI-XIIrHp_8:-L9HYuMKBvI-XIIrHp_9:b-cxdnq3) ([raw file](https://github.com/m-lab/mlab-ns/blob/74163e1b49547e2672ea90b2b3528ea6eee07415/server/mlabns/tests/test_prometheus_status.py#L89)):*
> ```Python
> 
>         urlopen_patch = mock.patch.object(
>             prometheus_status.urllib2.OpenerDirector,
> ```

I think I've done this before, but now believe it is discouraged to perform injection through other packages like this. Instead, it is preferable for this file (`test_prometheus_status.py`) to import `urllib2` and patch it on each function where necessary using the style described below. e.g. `@mock.patch.object(urllib2.OpenDirectory, 'open')`

---

*[server/mlabns/tests/test_prometheus_status.py, line 103 at r5](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9HUbcG43RflY0q0-wo:-L9HUbcG43RflY0q0-wp:b-7dqzxy) ([raw file](https://github.com/m-lab/mlab-ns/blob/74163e1b49547e2672ea90b2b3528ea6eee07415/server/mlabns/tests/test_prometheus_status.py#L103)):*
> ```Python
>         parse_sliver_tool_status_patch.start()
> 
>     def test_get_slice_status_returns_none_with_invalid_json(self):
> ```

Rather than starting the patch in `setUp`, I suggest using `@mock.patch.object` whenever possible.

```
@mock.patch.object(prometheus_status, 'parse_sliver_tool_status', autospec=True)
def test_get_slice_status_returns_none_with_invalid_json(self, mock_parse_sliver_tool_status):
    mock_parse_sliver_tool_status.return_value = 'whatever'
    ...
```

This form automatically creates a mock and passes it as an additional parameter to the function, that you can then setup as needed for that test. After the test runs, the mocked object is returned to what it was originally.

The decorator can be applied multiple times to the same function.

---

*[server/mlabns/util/prometheus_status.py, line 149 at r5](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9HQE2jEYKZz_vpa-Xb:-L9HQE2jEYKZz_vpa-Xc:b-7vv9w0) ([raw file](https://github.com/m-lab/mlab-ns/blob/74163e1b49547e2672ea90b2b3528ea6eee07415/server/mlabns/util/prometheus_status.py#L149)):*
> ```Python
>     """
>     # Make sure that the status is minimally correct.
>     if not 'value' in status or not 'experiment' in status[
> ```

This check is good.

---

*[server/mlabns/util/prometheus_status.py, line 183 at r5](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9HPEcc1yKE40LbFKQe:-L9HPEcc1yKE40LbFKQf:bs6z7hu) ([raw file](https://github.com/m-lab/mlab-ns/blob/74163e1b49547e2672ea90b2b3528ea6eee07415/server/mlabns/util/prometheus_status.py#L183)):*
> ```Python
>     tool_name = tool_id + address_family
>     if not tool_name in QUERIES:
>         logging.error('There is no Prometheus query for tool: %s, tool_name')
> ```

Please double check the quoting here.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-:-L9Hd0jV34Z9yV68RdOu:b-fj3qph)*

stephen-soltesz:
:lgtm: -- If you can eliminate `MockHTTPError`, please do so. Otherwise, this looks good!

---

Reviewed 2 of 2 files at r6.
Review status: all files reviewed at latest revision, all discussions resolved.

---

*[server/mlabns/tests/test_prometheus_status.py, line 90 at r6](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9M4OCP8GkbEEQr_LNo:-L9M4OCP8GkbEEQr_LNp:b-vnhiaa) ([raw file](https://github.com/m-lab/mlab-ns/blob/c7f5d1dad696db7d78730ad90d1f45374ceb2a37/server/mlabns/tests/test_prometheus_status.py#L90)):*
> ```Python
> class StatusUpdateHandlerTest(unittest2.TestCase):
> 
>     def setUp(self):
> ```

So, I was worried that this function was called once per class. But, it seems it's called before every function. So, TIL.

---

*[server/mlabns/tests/test_prometheus_status.py, line 156 at r6](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9M444mDIKXobCDf11a:-L9M444nAGEjykj36KnL:bpu1tud) ([raw file](https://github.com/m-lab/mlab-ns/blob/c7f5d1dad696db7d78730ad90d1f45374ceb2a37/server/mlabns/tests/test_prometheus_status.py#L156)):*
> ```Python
>             self, mock_open):
> 
>         class MockHttpError(urllib2.HTTPError):
> ```

Do we need this class? Can we create a literal `urllib2.HTTPError`? We don't check the self.cause or interact with the result in any way.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-:-L9M4qUG4eABGlXzovl_:bk4fk7o)*

nkinkade:
Review status: 10 of 11 files reviewed at latest revision, 2 unresolved discussions.

---

*[server/mlabns/tests/test_prometheus_status.py, line 156 at r6](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-L9M444mDIKXobCDf11a:-L9MFlZa81P8T_vu11eQ:bvg6urk) ([raw file](https://github.com/m-lab/mlab-ns/blob/c7f5d1dad696db7d78730ad90d1f45374ceb2a37/server/mlabns/tests/test_prometheus_status.py#L156)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

Do we need this class? Can we create a literal `urllib2.HTTPError`? We don't check the self.cause or interact with the result in any way.
</blockquote></details>

From Slack:

```
Nathan Kinkade: @soltesz: Do you know offhand of a good way to mock a `urllib2.HTTPError`?
Apparently that class is instantiated with 6 arguments, which seem excessive.
I'm probably going about this the wrong way.

Stephen Soltesz: Ooh. So, the mock class was to make it easier to create an instance... hmm. 6 arguments is excessive ..
Well if we're creating the mock class to override __init__ then saying that in a comment and leaving the current implementation would suffice
```

Leaving as is, but added a comment to explain the subclassing.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128)*

nkinkade:
:lgtm:

---

Review status: 10 of 11 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/m-lab/mlab-ns/128#-:-L9MGgT71lSYllu8_euT:bnfp4nl)*

PR #129, opened 2018-04-06
Adds a simple python script to compare two mlab-ns instance results.

nkinkade:
This simple python script queries two separate mlab-ns instances and then compares which services are advertised for a given experiment. A typical usage of this will be to compare results from a production instance of mlab-ns with results from a testing instance to be sure that they mostly align.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/129)

### DIFF
--- a/compare_instances.py
+++ b/compare_instances.py
@@ -264,7 +264,7 @@ def main():
             discrepancy_2_1_avg)
 
         # Sleep for 60s
-        time.sleep(10)
+        time.sleep(60)
 
 
 if __name__ == '__main__':
--- a/compare_instances.py
+++ b/compare_instances.py
@@ -4,7 +4,9 @@
 import argparse
 import json
 import logging
+import re
 import sys
+import time
 import urllib2
 
 QUERY_STRING = 'policy=all'
@@ -21,7 +23,7 @@ def parse_options(args):
         their values.
     """
     parser = argparse.ArgumentParser(
-        description='Compares the results from two different mlab-ns instances.')
+        description='Compares the results from two mlab-ns instances.')
     parser.add_argument(
         '--instance_one_url',
         dest='instance_one_url',
@@ -36,10 +38,21 @@ def parse_options(args):
                         dest='tool_id',
                         default='ndt',
                         help='The tool_id to query e.g., ndt.')
-    parser.add_argument('--all_fields',
-                        dest='all_fields',
-                        action="store_true",
-                        help='Whether to compare all fields or just "fqdn".')
+    parser.add_argument(
+        '--all_fields',
+        dest='all_fields',
+        action="store_true",
+        help='Compare all fields for equality, not just "fqdn".')
+    parser.add_argument(
+        '--infer_queueing',
+        dest='infer_queueing',
+        action="store_true",
+        help='Attempts to determine if queueing accounts for differences.')
+    parser.add_argument('--ignore_pattern',
+                        dest='ignore_patterns',
+                        nargs='*',
+                        default=[],
+                        help='Regex pattern to ignore, matched against "fqdn".')
 
     args = parser.parse_args(args)
 
@@ -71,9 +84,44 @@ def get_mlabns_status(url):
     return json_data
 
 
-def convert_sort_results(results, args):
+def filter_results(results, args):
+    """Filters results for any the user didn't want processed.
+    
+    Arguments:
+        results: list of dicts, with each dict containing status for a service.
+        args: argparse.ArgumentParser, options passed to script.
+    
+    Returns:
+        list: with any matching items filtered.
+    """
+    # Before processing the result list, check to see if the 'fqdn'
+    # field matches any --ignore_pattern options that may have been supplied by
+    # the user.
+    filtered_results = []
+    for key, status in enumerate(results):
+        matched = False
+        for pattern in args.ignore_patterns:
+            if re.search(pattern, status['fqdn']):
+                matched = True
+
+        if not matched:
+            filtered_results.append(results[key])
+
+    return filtered_results
+
+
+def convert_results(results, args):
+    """Converts JSON results into a python set().
+    
+    Arguments:
+        results: list of dicts, with each dict containing status for a service.
+        args: argparse.ArgumentParser, options passed to script.
+    
+    Returns:
+        set: representing the results passed to function.
+    """
     results_list = []
-    for result in sorted(results, key=lambda d: d['fqdn']):
+    for result in results:
         # TODO (kinkade): make this work for IPv6 too.
         # result['ip'] is a list with the first item being the IPv4 address and
         # the second the IPv6, if one exists.
@@ -85,40 +133,62 @@ def convert_sort_results(results, args):
         else:
             item = [result['fqdn']]
 
-        results_list.append(item)
+        results_list.append(tuple(item))
 
-    result_set = set(tuple(row) for row in results_list)
-    return result_set
+    return set(results_list)
 
 
 def compare_results(s1, s2):
-    s1_not_s2 = s1.difference(s2)
-    s2_not_s1 = s2.difference(s1)
+    """Compares two sets.
+    
+    Arguments:
+        s1: set, first set for comparison.
+        s2: set, first set for comparison.
+    
+    Returns:
+        dict, data with comparison metrics for sets.
+    """
+    data = {}
+    data['s1_count'] = len(s1)
+    data['s1_not_s2'] = len(s1.difference(s2))
+    data['s2_not_s1'] = len(s2.difference(s1))
+    data['s1_and_s2'] = len(s1.intersection(s2))
 
-    print 'Instance 1 and 2 agree on %d services.' % len(s1.intersection(s2))
-    print 'Instance 1 advertises these services which instance 2 does not:'
-    for item in sorted(s1_not_s2):
-        print '    %s' % (item,)
-
-    print 'Instance 2 advertises these services which instance 1 does not:'
-    for item in sorted(s2_not_s1):
-        print '    %s' % (item,)
+    return data
 
 
 def main():
     args = parse_options(sys.argv[1:])
 
-    url_one = '%s/%s?%s' % (args.instance_one_url, args.tool_id, QUERY_STRING)
-    results_one = get_mlabns_status(url_one)
-    logging.info('Results returned from instance one: %d', len(results_one))
-    set_one = convert_sort_results(results_one, args)
+    s1_not_s2_total = 0
+    s2_not_s1_total = 0
+    s1_and_s2_total = 0
 
-    url_two = '%s/%s?%s' % (args.instance_two_url, args.tool_id, QUERY_STRING)
-    results_two = get_mlabns_status(url_two)
-    logging.info('Results returned from instance two: %d', len(results_two))
-    set_two = convert_sort_results(results_two, args)
+    while True:
+        url_one = '%s/%s?%s' % (args.instance_one_url, args.tool_id,
+                                QUERY_STRING)
+        results_one = get_mlabns_status(url_one)
+        results_one_filtered = filter_results(results_one, args)
+        set_one = convert_results(results_one_filtered, args)
 
-    compare_results(set_one, set_two)
+        url_two = '%s/%s?%s' % (args.instance_two_url, args.tool_id,
+                                QUERY_STRING)
+        results_two = get_mlabns_status(url_two)
+        results_two_filtered = filter_results(results_two, args)
+        set_two = convert_results(results_two_filtered, args)
+
+        data = compare_results(set_one, set_two)
+        s1_and_s2_total = s1_and_s2_total + data['s1_and_s2']
+        s1_not_s2_total = s1_not_s2_total + data['s1_not_s2']
+        s2_not_s1_total = s2_not_s1_total + data['s2_not_s1']
+        print 'Instance 1 AND instance 2: %d' % data['s1_and_s2']
+        print 'Instance 1 NOT instance 2: %d' % data['s1_not_s2']
+        print 'Instance 2 NOT instance 1: %d' % data['s2_not_s1']
+        discrepancy = data['s1_not_s2'] / (data['s1_count'] * 1.0) * 100
+        print 'Discrepancy between 1 and 2: {0:.2f}%'.format(discrepancy)
+
+        # Sleep for 100s
+        time.sleep(100)
 
 
 if __name__ == '__main__':
--- a/compare_instances.py
+++ b/compare_instances.py
@@ -48,6 +48,10 @@ def parse_options(args):
         dest='infer_queueing',
         action="store_true",
         help='Attempts to determine if queueing accounts for differences.')
+    parser.add_argument('--verbose',
+                        dest='verbose',
+                        action="store_true",
+                        help='Whether to print extra information.')
     parser.add_argument('--ignore_pattern',
                         dest='ignore_patterns',
                         nargs='*',
@@ -110,6 +114,36 @@ def filter_results(results, args):
     return filtered_results
 
 
+def infer_queueing(results1, results2):
+    """Makes a loose determination of whether NDT node is queueing.
+
+    NOTE: This function is temporary and highly specific to comparing Nagios status to Prometheus status, AND is only relevant for --tool_id=ndt. Nagios does not know whether NDT is queueing or not, whereas Prometheus has a check for this, and will fail NDT if it detects that queueing is happening. This function dumbly assumes that Nagios (results1) is advertising mlab1, and if Prometheus (results2) is advertising anything other than mlab1 for the same site, then we assume it has detected queueing.
+
+    Arguments:
+        results1: list of dicts, with each dict containing status for a service.
+        results2: list of dicts, with each dict containing status for a service.
+
+    Returns:
+        list, list of dicts with possible NDT queueing reconciled.
+    """
+    munged_results2 = []
+    for result1 in results1:
+        r1_node, r1_site = re.search('(mlab[1-4])\.([a-z]{3}[0-9]{2})',
+                                     result1['fqdn']).groups()
+        for result2 in results2:
+            r2_node, r2_site = re.search('(mlab[1-4])\.([a-z]{3}[0-9]{2})',
+                                         result2['fqdn']).groups()
+            if r1_site == r2_site:
+                if r1_node < r2_node:
+                    munged_results2.append(result1)
+                else:
+                    munged_results2.append(result2)
+            else:
+                munged_results2.append(result2)
+
+    return munged_results2
+
+
 def convert_results(results, args):
     """Converts JSON results into a python set().
     
@@ -146,13 +180,12 @@ def compare_results(s1, s2):
         s2: set, first set for comparison.
     
     Returns:
-        dict, data with comparison metrics for sets.
+        dict, data with comparison results for sets.
     """
     data = {}
-    data['s1_count'] = len(s1)
-    data['s1_not_s2'] = len(s1.difference(s2))
-    data['s2_not_s1'] = len(s2.difference(s1))
-    data['s1_and_s2'] = len(s1.intersection(s2))
+    data['s1_not_s2'] = s1.difference(s2)
+    data['s2_not_s1'] = s2.difference(s1)
+    data['s1_and_s2'] = s1.intersection(s2)
 
     return data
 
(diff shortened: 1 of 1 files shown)


stephen-soltesz:

---

*[compare_instances.py, line 76 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9RS_gL7s4h0ADX2ex-:-L9RS_gMCcpeWxh27H-R:b2t9nyk) ([raw file](https://github.com/m-lab/mlab-ns/blob/e25dd60fe18ad6f190fb925200a0f93602bdeff6/compare_instances.py#L76)):*
> ```Python
> def convert_sort_results(results, args):
>     results_list = []
>     for result in sorted(results, key=lambda d: d['fqdn']):
> ```

Is `site` a better key to sort by? Or, does order matter at all if we produce a `set()` below.

---

*[compare_instances.py, line 90 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9RT66bDp-BRifUrWqY:-L9RT66bDp-BRifUrWqZ:b-7slgr8) ([raw file](https://github.com/m-lab/mlab-ns/blob/e25dd60fe18ad6f190fb925200a0f93602bdeff6/compare_instances.py#L90)):*
> ```Python
>         results_list.append(item)
> 
>     result_set = set(tuple(row) for row in results_list)
> ```

If you create the tuple above, when assigning `item`, this could simply read `set(results_list)`.

---

*[compare_instances.py, line 121 at r1](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9RToch6ABewWCD39zk:-L9RToch6ABewWCD39zl:bcvw8zq) ([raw file](https://github.com/m-lab/mlab-ns/blob/e25dd60fe18ad6f190fb925200a0f93602bdeff6/compare_instances.py#L121)):*
> ```Python
>     set_two = convert_sort_results(results_two, args)
> 
>     compare_results(set_one, set_two)
> ```

Consider running the "fetch and compare" logic in a loop, that runs once a minute.

Also, I think of mlab-ns as being the authoritative answer. And, we want to measure how often mlab-nstesting is different. I think we want to talk about the percent of records in nstesting that were: missing, extra, in-common wrt mlab-ns.

So, what if `compare_results` returned counts for: missing_from_s2_count, extra_in_s2_count, in_common_count. And, if there were an additional parameter for total_number_of_servers, we would have a denominator to calculate a percentages.

The loop could also keep a cumulative sum of these counts, and the number of rounds, so we can calculate the percentage of missing, extra, and matching over time. Ideally the missing and extra numbers should be low, like less than 1%.

And, if we ran that for an hour or something, we could be pretty comfortable that the two systems are behaving equally to one another.

The loop could print two lines, the instantaneous percentage and the current cumulative percentage for these values.

How does this sound?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129)*

stephen-soltesz:
Great. I've added more notes on the calculations that I think will help us find the percentage of time the two services agree or disagree.

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[compare_instances.py, line 130 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9fZkmnEFGqEXEjU363:-L9fZkmnEFGqEXEjU364:b4rwbpr) ([raw file](https://github.com/m-lab/mlab-ns/blob/36cfbc0bea50280169a2e0fed2d22eb979abd1f4/compare_instances.py#L130)):*
> ```Python
>     """
>     munged_results2 = []
>     for result1 in results1:
> ```

I realize this is a hack. If this were not a hack, I would ask for a bigger rewrite, because we could do this more efficiently.

---

*[compare_instances.py, line 136 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9fZlPEEzQjaH3NZK1U:-L9fZlPEEzQjaH3NZK1V:b-mnsqp5) ([raw file](https://github.com/m-lab/mlab-ns/blob/36cfbc0bea50280169a2e0fed2d22eb979abd1f4/compare_instances.py#L136)):*
> ```Python
>             r2_node, r2_site = re.search('(mlab[1-4])\.([a-z]{3}[0-9]{2})',
>                                          result2['fqdn']).groups()
>             if r1_site == r2_site:
> ```

Please combine the condition `r1_site == r2_site and r1_host < r2_host` to make the branch logic simpler.

---

*[compare_instances.py, line 202 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9fwtMHBnkeBKbCv_zV:-L9fwtMHBnkeBKbCv_zW:b-m4ksgx) ([raw file](https://github.com/m-lab/mlab-ns/blob/36cfbc0bea50280169a2e0fed2d22eb979abd1f4/compare_instances.py#L202)):*
> ```Python
>     s2_not_s1_total = 0
> 
>     while samples < 10:
> ```

Please move this loop to it's own function. `main` is getting pretty big.

---

*[compare_instances.py, line 225 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9fcfMTDzKVWxsBFVKZ:-L9fcfMTDzKVWxsBFVK_:bo8g7fc) ([raw file](https://github.com/m-lab/mlab-ns/blob/36cfbc0bea50280169a2e0fed2d22eb979abd1f4/compare_instances.py#L225)):*
> ```Python
> 
>         # Calculate counts and totals
>         s1_count = len(set_one)
> ```

I don't think we want to use s1_count in the calculations below.

Since mlab-ns only returns "online" results, we cannot calculate the absolute percentage of hosts online without external input. Also, it needs to be verified independently, that the total number known to Nagios & Prometheus are equal.

Adding a new parameter was my original suggestion but I think we can do better with the data we have.

Whatever the total absolute number of hosts, the counts should always satisfy the equation:
```
total = (online in common) + (online missing) + (online extra) + (offline in common)
```
(Please confirm that this equation counts all elements exactly once.)

Using N for the set of online hosts from Nagios, P the set of online hosts from Prometheus, then the following set operations define each term as:
```
(online in common count) = len(N & P)
(online missing count) = len(N - P)
(online extra count) = len(P - N)
(offline in common count) = total - len(P || N)
```

We can rearrange terms to have:
```
len(P||N) = len(N&P) + [len(N-P) + len(P-N)]
```

If we divide both sides by `len(P||N)` then we get a sum of ratios that look like: %agree + %disgree.

```
% agree = len(N&P) / len(P||N)
% disagree = [len(N-P) + len(P-N)] / len(P||N)
```

---

*[compare_instances.py, line 226 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9fatRO2ZLPxite04z4:-L9fatRO2ZLPxite04z5:b-boqrqd) ([raw file](https://github.com/m-lab/mlab-ns/blob/36cfbc0bea50280169a2e0fed2d22eb979abd1f4/compare_instances.py#L226)):*
> ```Python
>         # Calculate counts and totals
>         s1_count = len(set_one)
>         s1_and_s2_count = len(data['s1_and_s2'])
> ```

In my head, I'm converting `s1_and_s2` to "in common" and `s1_not_s2` to "missing", etc. I find the current names much harder to reason about. Would you be willing to use the friendlier names?

---

*[compare_instances.py, line 235 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9fcA1_4jyN3A_e_9NA:-L9fcA1_4jyN3A_e_9NB:b5kegrm) ([raw file](https://github.com/m-lab/mlab-ns/blob/36cfbc0bea50280169a2e0fed2d22eb979abd1f4/compare_instances.py#L235)):*
> ```Python
> 
>         # Current
>         print '## SAMPLE %d' % samples
> ```

I suggest splitting this data into columns, so that every line is a new sample.

Imagine a CSV header and rows. But, just using spaces is fine too.

```
sample,union_count,common_count,missing_count,extra_count,pct_agree,pct_disagree,avg_*
1,244,240,1,3,0.016,...
2,...
```

With output in this form, we can easily run the script for any length of time, and then plot any of the columns over time. Ideally, discrepancy would always be zero, but realistically, we'll want to keep it below some threshold. Discovering what the typical discrepancy is today would be helpful for establishing a baseline.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-:-L9fxbKK3AWPbIbIPIls:b-5mrepq)*

nkinkade:
Thanks for the feedback. I'm implemented most of the changes you suggested. PTAL?

---

Review status: 0 of 1 files reviewed at latest revision, 7 unresolved discussions.

---

*[compare_instances.py, line 130 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9fZkmnEFGqEXEjU363:-L9fzA6s8NL2xPELWB9m:b-lfzvh2) ([raw file](https://github.com/m-lab/mlab-ns/blob/36cfbc0bea50280169a2e0fed2d22eb979abd1f4/compare_instances.py#L130)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

I realize this is a hack. If this were not a hack, I would ask for a bigger rewrite, because we could do this more efficiently.
</blockquote></details>

This could be a learning opportunity for me. If it doesn't take too much time, would you be willing to give me some insight into the ideas you have about making this a lot more simple? What I have done, to fix a bug, is to removed the 2nd `for` loop and replace it with a list comprehension that extracts the value I want. What more could/should I be doing?

---

*[compare_instances.py, line 136 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9fZlPEEzQjaH3NZK1U:-L9g-CPA3ZuZdwqi2X4S:b-896fix) ([raw file](https://github.com/m-lab/mlab-ns/blob/36cfbc0bea50280169a2e0fed2d22eb979abd1f4/compare_instances.py#L136)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

Please combine the condition `r1_site == r2_site and r1_host < r2_host` to make the branch logic simpler.
</blockquote></details>

Done.

---

*[compare_instances.py, line 202 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9fwtMHBnkeBKbCv_zV:-L9gqCsj7ejudf3C98xC:b-7ggafx) ([raw file](https://github.com/m-lab/mlab-ns/blob/36cfbc0bea50280169a2e0fed2d22eb979abd1f4/compare_instances.py#L202)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

Please move this loop to it's own function. `main` is getting pretty big.
</blockquote></details>

I haven't done this step yet. Doing this will make `main` about 5 lines long, and push all the ugliness into some other place. It will also require pushing variables around as function parameters a bit more. Is it fine form to push ugly thing out of `main` into a function right above `main`? If so, I can do that.

---

*[compare_instances.py, line 225 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9fcfMTDzKVWxsBFVKZ:-L9giI1zDvu2QjeIySE6:b-scu4z0) ([raw file](https://github.com/m-lab/mlab-ns/blob/36cfbc0bea50280169a2e0fed2d22eb979abd1f4/compare_instances.py#L225)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

I don't think we want to use s1_count in the calculations below.

Since mlab-ns only returns "online" results, we cannot calculate the absolute percentage of hosts online without external input. Also, it needs to be verified independently, that the total number known to Nagios & Prometheus are equal.

Adding a new parameter was my original suggestion but I think we can do better with the data we have.

Whatever the total absolute number of hosts, the counts should always satisfy the equation:
```
total = (online in common) + (online missing) + (online extra) + (offline in common)
```
(Please confirm that this equation counts all elements exactly once.)

Using N for the set of online hosts from Nagios, P the set of online hosts from Prometheus, then the following set operations define each term as:
```
(online in common count) = len(N & P)
(online missing count) = len(N - P)
(online extra count) = len(P - N)
(offline in common count) = total - len(P || N)
```

We can rearrange terms to have:
```
len(P||N) = len(N&P) + [len(N-P) + len(P-N)]
```

If we divide both sides by `len(P||N)` then we get a sum of ratios that look like: %agree + %disgree.

```
% agree = len(N&P) / len(P||N)
% disagree = [len(N-P) + len(P-N)] / len(P||N)
```
</blockquote></details>

I think this is implemented now.

---

*[compare_instances.py, line 226 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9fatRO2ZLPxite04z4:-L9ghvCeBA7yGC1u-3d1:b-xq9lyp) ([raw file](https://github.com/m-lab/mlab-ns/blob/36cfbc0bea50280169a2e0fed2d22eb979abd1f4/compare_instances.py#L226)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

In my head, I'm converting `s1_and_s2` to "in common" and `s1_not_s2` to "missing", etc. I find the current names much harder to reason about. Would you be willing to use the friendlier names?
</blockquote></details>

Done.  I had been using these semi obscure names because I wasn't sure that we'd always be using this with the assumption that s1 was the authoritative source on which to compare everything else. But it we assume this is the case, then applying the label "missing" to entries that are in s1 but not s2 makes more sense.

---

*[compare_instances.py, line 235 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9fcA1_4jyN3A_e_9NA:-L9gq5Ie02jqtlPFgehz:b-896fix) ([raw file](https://github.com/m-lab/mlab-ns/blob/36cfbc0bea50280169a2e0fed2d22eb979abd1f4/compare_instances.py#L235)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

I suggest splitting this data into columns, so that every line is a new sample.

Imagine a CSV header and rows. But, just using spaces is fine too.

```
sample,union_count,common_count,missing_count,extra_count,pct_agree,pct_disagree,avg_*
1,244,240,1,3,0.016,...
2,...
```

With output in this form, we can easily run the script for any length of time, and then plot any of the columns over time. Ideally, discrepancy would always be zero, but realistically, we'll want to keep it below some threshold. Discovering what the typical discrepancy is today would be helpful for establishing a baseline.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-:-L9gqZDz6zmp18lhuZrm:b-5rjvji)*

stephen-soltesz:
:lgtm: -- So, what does it report?

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[compare_instances.py, line 130 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9fZkmnEFGqEXEjU363:-L9k9Ut30A3N--P42V4H:b-z84il2) ([raw file](https://github.com/m-lab/mlab-ns/blob/36cfbc0bea50280169a2e0fed2d22eb979abd1f4/compare_instances.py#L130)):*

This could be a learning opportunity for me. If it doesn't take too much time, would you be willing to give me some insight into the ideas you have about making this a lot more simple? What I have done, to fix a bug, is to removed the 2nd `for` loop and replace it with a list comprehension that extracts the value I want. What more could/should I be doing?
</blockquote></details>

The loop within a loop is N^2. If the inputs were dicts, say fqdn to "result", then the lookups could be much faster. For just 500 records maybe this doesn't matter. Just seeing N^2 loops makes me squint suspiciously.

---

*[compare_instances.py, line 202 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9fwtMHBnkeBKbCv_zV:-L9kQs7l8gXDouJgzpZh:b-z703b3) ([raw file](https://github.com/m-lab/mlab-ns/blob/36cfbc0bea50280169a2e0fed2d22eb979abd1f4/compare_instances.py#L202)):*

I haven't done this step yet. Doing this will make `main` about 5 lines long, and push all the ugliness into some other place. It will also require pushing variables around as function parameters a bit more. Is it fine form to push ugly thing out of `main` into a function right above `main`? If so, I can do that.
</blockquote></details>

¯\_(ツ)_/¯ -- imo, it would be better.

---

*[compare_instances.py, line 243 at r4](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9k7vRF8wozh9WjzI-V:-L9k7vRF8wozh9WjzI-W:b-ymfbn6) ([raw file](https://github.com/m-lab/mlab-ns/blob/b2ca758a1f0bec365ddc4d93ae67ce0ef079a5d1/compare_instances.py#L243)):*
> ```Python
> 
>         # Calculate average counts and totals
>         common_total = common_total + len(data['common'])
> ```

`+=` reads a little more easily.

```
common_total += len(data['common'])
```

---

*[compare_instances.py, line 256 at r4](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9k8HdZ6n3GWosJzgQW:-L9k8HdZ6n3GWosJzgQX:bgi4t0x) ([raw file](https://github.com/m-lab/mlab-ns/blob/b2ca758a1f0bec365ddc4d93ae67ce0ef079a5d1/compare_instances.py#L256)):*
> ```Python
>             pct_disagree=(
>                 (missing_count + extra_count) / (denominator * 1.0) * 100),
>             avg_common=(common_total / samples),
> ```

Real question: do we want integer division for the avgs?

---

*[compare_instances.py, line 260 at r4](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9k8lM21auwSv-kSNWZ:-L9k8lM21auwSv-kSNW_:b-aq9whc) ([raw file](https://github.com/m-lab/mlab-ns/blob/b2ca758a1f0bec365ddc4d93ae67ce0ef079a5d1/compare_instances.py#L260)):*
> ```Python
>             avg_extra=(extra_total / samples),
>             avg_pct_agree=((common_total / samples) /
>                            (denominator_total / samples * 1.0) * 100),
> ```

The expression can be simplified: (a/c) / (b/c) == (a / b)

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-:-L9kRFHUB6c8qZEr8Jl-:b-ygccny)*

nkinkade:
With `--infer_queueing` enabled and `--ignore_pattern 'tyo'` it seems to report a very low, sometimes even 0, difference. I'm going to run it for an hour after I merge this. Thanks!

---

Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions.

---

*[compare_instances.py, line 130 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9fZkmnEFGqEXEjU363:-L9kat3yESOXOAg9-SNQ:bvhawkz) ([raw file](https://github.com/m-lab/mlab-ns/blob/36cfbc0bea50280169a2e0fed2d22eb979abd1f4/compare_instances.py#L130)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

The loop within a loop is N^2. If the inputs were dicts, say fqdn to "result", then the lookups could be much faster. For just 500 records maybe this doesn't matter. Just seeing N^2 loops makes me squint suspiciously.
</blockquote></details>

Is the list comprehension I added an improvement in your eyes? It seems to be, at least to my eye.

---

*[compare_instances.py, line 202 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9fwtMHBnkeBKbCv_zV:-L9kazW81BK9IvI9OerG:bot0hxx) ([raw file](https://github.com/m-lab/mlab-ns/blob/36cfbc0bea50280169a2e0fed2d22eb979abd1f4/compare_instances.py#L202)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

¯\_(ツ)_/¯ -- imo, it would be better.
</blockquote></details>

I'll move it in deference to your experience and attention to best form.

---

*[compare_instances.py, line 243 at r4](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9k7vRF8wozh9WjzI-V:-L9k_W8lCk48hsbHsxL2:bpt7ddv) ([raw file](https://github.com/m-lab/mlab-ns/blob/b2ca758a1f0bec365ddc4d93ae67ce0ef079a5d1/compare_instances.py#L243)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

`+=` reads a little more easily.

```
common_total += len(data['common'])
```
</blockquote></details>

It is neater and shorter and I'll make the change. In this case I made a conscious decision not to use the `+=` operator because of a memory 
 I have from reading [PEP 20](https://www.python.org/dev/peps/pep-0020/#id3) that states that explicit is *always* better than implicit, and I have some foggy memory that somewhere I someone once made the example of the difference between `x = x + n` vs `x += n`, indicating that the former, while longer, was perhaps preferable in many cases because it is more explicit  and obvious.  That said, my memory may be failing me, and it might be that  the "pythonic" way of doing things is `+=` whenever you can.

---

*[compare_instances.py, line 256 at r4](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9k8HdZ6n3GWosJzgQW:-L9kqqgr7dvY4cVfanTs:b-pz1fpb) ([raw file](https://github.com/m-lab/mlab-ns/blob/b2ca758a1f0bec365ddc4d93ae67ce0ef079a5d1/compare_instances.py#L256)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

Real question: do we want integer division for the avgs?
</blockquote></details>

I don't know. But to get floating point division, I've added `from __future__ import division` to the top of the file, and have removed the awkward multiplication by 1.0 to get floating point results. It is now fp without that.

---

*[compare_instances.py, line 260 at r4](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9k8lM21auwSv-kSNWZ:-L9kbmKh67M0hG26KKi8:b5wqnw1) ([raw file](https://github.com/m-lab/mlab-ns/blob/b2ca758a1f0bec365ddc4d93ae67ce0ef079a5d1/compare_instances.py#L260)):*
<details><summary><i>Previously, stephen-soltesz (Stephen Soltesz) wrote…</i></summary><blockquote>

The expression can be simplified: (a/c) / (b/c) == (a / b)
</blockquote></details>

Thank you. Fixed. It's cases like this where my lack of math aptitude and experience shows.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-:-L9kt6aZ75eqS9DWD5pK:b-lyjb06)*

nkinkade:
:lgtm: Adding this myself, since Github doesn't seem to recognize a `:lgtm:` from Reviewable if it's buried one or two comments in.

---

Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-:-L9kth6Y97NJRo1ESAI7:bkq6af4)*

stephen-soltesz:
Cool!

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[compare_instances.py, line 130 at r3](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9fZkmnEFGqEXEjU363:-L9kve1d6KX2ONKuP6f1:bz0vpx0) ([raw file](https://github.com/m-lab/mlab-ns/blob/36cfbc0bea50280169a2e0fed2d22eb979abd1f4/compare_instances.py#L130)):*

Is the list comprehension I added an improvement in your eyes? It seems to be, at least to my eye.
</blockquote></details>

The list comprehension makes the logic more compact and readable. However, the list comprehension is just an abbreviated form of the inner `for` loop. So the overall runtime remains N^2.

This specific thing is not that big a deal.

I know this script is basically a one-time (or few time) use which means we shouldn't spend too much time on it. But, leaving things I would ordinarily not let go makes me twitchy. And, that I commented on this rather than other things doesn't mean as much as it seems other than I'm feeling twitchy.

---

*[compare_instances.py, line 243 at r4](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-L9k7vRF8wozh9WjzI-V:-L9kuaKg4e3Fy3Ruz6qr:b95xpf0) ([raw file](https://github.com/m-lab/mlab-ns/blob/b2ca758a1f0bec365ddc4d93ae67ce0ef079a5d1/compare_instances.py#L243)):*

It is neater and shorter and I'll make the change. In this case I made a conscious decision not to use the `+=` operator because of a memory 
 I have from reading [PEP 20](https://www.python.org/dev/peps/pep-0020/#id3) that states that explicit is *always* better than implicit, and I have some foggy memory that somewhere I someone once made the example of the difference between `x = x + n` vs `x += n`, indicating that the former, while longer, was perhaps preferable in many cases because it is more explicit  and obvious.  That said, my memory may be failing me, and it might be that  the "pythonic" way of doing things is `+=` whenever you can.
</blockquote></details>

"Explicit is better than implicit" is somewhat ironic in the context of Python, where there can be so much object magic happening and `__future__` import changes that it can be hard to know what's happening under the surface.

AFAIK, `+=` is good form in Python. The https://google.github.io/styleguide/pyguide.html does not provide an opinion other than to say *don't* use it for strings, which sounds right.

If you do find another source that makes the case against `+=` please let me know.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/m-lab/mlab-ns/129#-:-L9kwVnF2fp1twiUgNqf:b-89q77x)*

PR #135, opened 2018-05-08
fix roundrobin bug

yachang:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/135)

### DIFF
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -119,6 +119,9 @@ class SiteRegistrationHandler(webapp.RequestHandler):
                                   site[self.SITE_FIELD])
                     continue
 
+        # call check_ip job at the end of check_site job
+        IPUpdateHandler().get()
+
         return util.send_success(self)
 
     def update_site(self, site):


nkinkade:

---

*[server/mlabns/handlers/update.py, line 123 at r2](https://reviewable.io/reviews/m-lab/mlab-ns/135#-LC-KqWtD_e3Kky_kkjs:-LC-KqWtD_e3Kky_kkjt:b-4nz86n) ([raw file](https://github.com/m-lab/mlab-ns/blob/4326e1c1dadb3da131dd8fe6296b129536739c2f/server/mlabns/handlers/update.py#L123)):*
> ```Python
> 
>         # call check_ip job at the end of check_site job
>         IPUpdateHandler().get()
> ```

Running the IPUpdateHanlder() as part of `check_site` is probably fine, but if we do this, then we will surely want to disable the `check_ip` cron job.

Another concern I have with this is that the combined run time of `SiteRegistrationHandler()` and `IPUpdateHandler()` is currently going to be around 10 minutes. Meanwhile the `check_status` cron job will have run about 10 times, reading from memcache and unconditionally overwriting it once every minute. I worry that it will likely overwrite any updates to memcache by this new long-running cron job. What are your thoughts on this?

---

*[server/mlabns/handlers/update.py, line 174 at r2](https://reviewable.io/reviews/m-lab/mlab-ns/135#-LC-Lnzj5W3BM7NzaqK8:-LC-Lnzj5W3BM7NzaqK9:b2owese) ([raw file](https://github.com/m-lab/mlab-ns/blob/4326e1c1dadb3da131dd8fe6296b129536739c2f/server/mlabns/handlers/update.py#L174)):*
> ```Python
>                 if not memcache.set(
>                         tool.tool_id,
>                         sliver_tool,
> ```

Will this overwrite the same memcache key (`tool_id`, e.g., ndt, neubot, etc.) three different times with different data? `tool_id` [comes from an outer loop](https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/handlers/update.py#L159), in which `tool.tool_id` will not change in the inner loop. The inner loop iterates over `['mlab1', 'mlab2', 'mlab3']`.

The other possible issue I see here is that this is a very large departure from my understanding of how memcache is currently used. This addition seems to be storing a single `sliver_tool` in memcache, keyed on `tool_id`. However, in other places the memcache key `tool_id` stores a large python object (a dict or list of dicts) and the code expects, for example, that something like `memcache.get('ndt')` is going to return every single NDT sliver tool, not just one single sliver tool. Examples:

https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/db/sliver_tool_fetcher.py#L200
https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/handlers/update.py#L509
https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/handlers/update.py#L279

I believe this is why the memcache viewer in the GCP console doesn't return anything readable, because there are only a small handful of memcache keys, one for each tool (ndt, ndt_ssl, neubot, mobiperf, npad, ooni), and behind that key is every single sliver_tool for that given tool.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/135)*

PR #137, opened 2018-05-08
Use MaxMind GeoLite2 databases instead of the deprecated "legacy" ones.

nkinkade:
Currently, we are using an outdated, deprecated, legacy MaxMind database that will not receive many updates going forward, if any, and will [completely removed from MaxMind's website on January 2, 2019](https://support.maxmind.com/geolite-legacy-discontinuation-notice/).

This PR implements the changes necessary to start using the current [GeoLite2 databases](https://dev.maxmind.com/geoip/geoip2/geolite2/).

**NOTE**: This PR is not at all as large as it seems. All changes in the `mlabns/third_party/*` directory, which is the overwhelming vast majority of apparent changes in this PR, can be ignored, since it is all purely 3rd party vendor code fetched by pip. However, it should be noted that there *are* changes to the `geoip2` code, for which I'll be submitting a PR to MaxMind for, and is separate from this PR.

Of note in the PR:
* The MaxMind database file is no longer included in this repository. GAE has a limit of 32MB for static code files, and the new GeoLite2 database is 55MB. Instead, the file is stored in GCS and fetched from there (location of this file is in `util.constants`.
* The new GeoLite2 database contain information for both IPv4 and IPv6 lookups, thus we no longer need to distinguish between the two when doing lookups.
* There is a new `util.handler` class which will allow us to update the running version of the MaxMind database by simply fetching a URL from the GAE app (`/reload_maxminddb`).

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/137)


yachang:
This is a gigantic cl with 210 files!

Anyway to split it into several PRs, so you can find multiple reviewers, and speed up the process?

For example, make the 3rd pary code change a separtate PR? 

---

Review status: 0 of 210 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/137#-:-LC9sJDx73NSCZ-Hscpu:b-3rljo9)*

nkinkade:
@yachang: I could split out the `third_party/` files into a separate PR, but it seems just as easy to say: ignore any changes to files in `third_party/`. Excluding those vendor files, this PR only have relatively minimal changes to 5 or 6 mlab-ns files.  It seems to me this is more efficient, and it keeps like changes in a single PR. I could break all the `third_party/` files out into a separate PR for you and then just say: Ignore every single file/change in this entire PR. Or I can leave it in this PR and say: Ignore every single change `third_party/`. I don't see much benefit in a separate PR. But I may not be thinking about it correctly. If you disagree, could you elaborate?

---

Review status: 0 of 210 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[server/mlabns/tests/test_maxmind.py, line 98 at r3](https://reviewable.io/reviews/m-lab/mlab-ns/137#-LC0y1QoOKE8AQuW6fiX-r3-98:-LCEvb5m-BkShv4nNAi4:b-xzyjqj) ([raw file](https://github.com/m-lab/mlab-ns/blob/66e7c6b50e679e75a33fbf796c2ec7eedb416681/server/mlabns/tests/test_maxmind.py#L98)):*

Do you have a unittest to replace testGetGeolocationValidLocation()?
</blockquote></details>

I don't, but for a very specific reason which I address in a comment at the top of this class (`ModelMockup`). Please take a look at that comment and see if you perhaps disagree?

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/137#-:-LCEvtpJDyyXFq5jWgcA:bhxn2vi)*

nkinkade:
@yachang: Yes, I think I saw your comments. The only ones I saw were the one above about this being a gigantic PR, and the one below about:

> Do you have a unittest to replace testGetGeolocationValidLocation?

I answered both.  Are you not seeing my replies somehow?

---

Review status: 0 of 210 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/137#-:-LCF_YKS9bv_cV2Rq84N:bbayxfi)*

nkinkade:
@yachang: I think I've addressed your comment and concerns. PTAL?

---

Review status: 0 of 14 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[server/app.yaml, line 4 at r4](https://reviewable.io/reviews/m-lab/mlab-ns/137#-LC0y1QrRYyLyoa4LhLx-r4-4:-LCUhr_kE0cBGM_cRtCd:bwoo150) ([raw file](https://github.com/m-lab/mlab-ns/blob/be500240a8d258e36184ec82df70e53408e24c2f/server/app.yaml#L4)):*

Resolve with master branch. I have modified that in the roundrobin PR.
</blockquote></details>

I've removed this change from this PR.

---

*[server/mlabns/tests/test_maxmind.py, line 62 at r3](https://reviewable.io/reviews/m-lab/mlab-ns/137#-LC0y1QoOKE8AQuW6fiX-r3-62:-LCUha4C4QqgEYbnksLn:b-gut99s) ([raw file](https://github.com/m-lab/mlab-ns/blob/66e7c6b50e679e75a33fbf796c2ec7eedb416681/server/mlabns/tests/test_maxmind.py#L62)):*

Thanks for the explanation. Can you add a TODO for "create a fake database (same format but much smaller, maybe just several K), then check the validity of an entry using the fake database" ? Thanks!
</blockquote></details>

Done. I've add a new TODO.

---

*[server/mlabns/util/constants.py, line 23 at r3](https://reviewable.io/reviews/m-lab/mlab-ns/137#-LC0y1QoOKE8AQuW6fiZ-r3-23:-LCUh-6L36k9DlOVLwh5:bdkbqwr) ([raw file](https://github.com/m-lab/mlab-ns/blob/66e7c6b50e679e75a33fbf796c2ec7eedb416681/server/mlabns/util/constants.py#L23)):*

Which project does the bucket belong to? Could you add documentation for that?
</blockquote></details>

I'm hesitant to document which GCP project owns the bucket since the bucket name is going to change, and when it changes someone will need to remember to also update the note, and they may get out of sync. Also, because GCS bucket names are global (across every project), which project owns it may be less relevant. I've added a comment on how to use `gsutil` to determine which GCP project owns the bucket. Does that address your concern?

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/137#-:-LCUhwKLEo8kpd8W3d_5:bxsk8t3)*

yachang:
:lgtm:

---

Review status: 0 of 14 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/137#-:-LCUnuaE1DFWlRp3vWTg:bnfp4nl)*

PR #138, opened 2018-05-09
merge crone job check_site and check_ip into one

yachang:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/138)

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,6 @@ After the Datastore is populated with seed information, manually kick off the cr
 Run the following jobs from GCP under Compute > App Engine > Task queues > Cron Jobs.
 
 1. `/cron/check_site`
-1. `/cron/check_ip`
 1. `/cron/check_status`
 
 If bootstrapping was successful, you should see a populated map at the root mlab-ns URL (e.g. mlab-nstesting.appspot.com) with M-Lab's sites properly located.
--- a/server/app.yaml
+++ b/server/app.yaml
@@ -1,7 +1,7 @@
 runtime: python27
 api_version: 1
 threadsafe: false
-instance_class: F4
+instance_class: F4_1G
 
 handlers:
 
--- a/server/cron.yaml
+++ b/server/cron.yaml
@@ -10,4 +10,4 @@ cron:
 # Check for new sites every 24 hours, starting at 00:00
 - description: Check sites
   url: /cron/check_site
-  schedule: every day 00:00
+  schedule: every day 06:00
--- a/server/cron.yaml
+++ b/server/cron.yaml
@@ -11,10 +11,3 @@ cron:
 - description: Check sites
   url: /cron/check_site
   schedule: every day 00:00
-
-# Check for new/modified IP addresses every 24 hours, starting at 01:00
-# (this will update also IP addresses for new sites, if added)
-- description: Check IP addresses
-  url: /cron/check_ip
-  schedule: every day 01:00
-
--- a/server/cron.yaml
+++ b/server/cron.yaml
@@ -7,7 +7,8 @@ cron:
   url: /cron/check_status
   schedule: every 1 minutes
 
-# Check for new sites every 24 hours, starting at 00:00
+# checking for new sites, and for new and/or updated IP addresses, roundrobin information.
+# run every 24 hours, starting at 06:00
 - description: Check sites
   url: /cron/check_site
   schedule: every day 06:00
--- a/server/cron.yaml
+++ b/server/cron.yaml
@@ -7,8 +7,8 @@ cron:
   url: /cron/check_status
   schedule: every 1 minutes
 
-# checking for new sites, and for new and/or updated IP addresses, roundrobin information.
+# check for new sites, and for new and/or updated IP addresses, roundrobin information.
 # run every 24 hours, starting at 06:00
-- description: Check sites
+- description: Check sites, update IP and roundrobin
   url: /cron/check_site
   schedule: every day 06:00
--- a/server/main.py
+++ b/server/main.py
@@ -12,7 +12,6 @@ app = webapp.WSGIApplication(
     [(r'/', admin.AdminHandler),
     (r'/admin.*', admin.AdminHandler),
     (r'/cron/check_status', update.StatusUpdateHandler),
-    (r'/cron/check_ip', update.IPUpdateHandler),
     (r'/cron/check_site', update.SiteRegistrationHandler),
     (r'/privacy', privacy.PrivacyHandler),
     (r'/docs', docs.DocsHandler),
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -160,15 +160,11 @@ class SiteRegistrationHandler(webapp.RequestHandler):
         return True
 
 
-class IPUpdateHandler(webapp.RequestHandler):
+class IPUpdateHandler():
     """Updates SliverTools' IP addresses."""
 
     # TODO: There should eventually be a TESTING_IP_LIST_URL for testing purpose.
     IP_LIST_URL = 'https://storage.googleapis.com/operator-mlab-oti/metadata/v0/current/mlab-host-ips.txt'
-
-    def get(self):
-        self.update()
-        return util.send_success(self)
 
     def update(self):
         """Triggers the update handler.
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -250,7 +250,7 @@ class IPUpdateHandler(webapp.RequestHandler):
                 updated_sliver_tool = self.set_sliver_tool_ips(sliver_tool,
                                                                ipv4, ipv6)
                 # Update all sliver tool.
-                self.put_sliver_tool(updated_sliver_tool)
+                #self.put_sliver_tool(updated_sliver_tool)
 
                 if tool.tool_id not in sliver_tool_list:
                     sliver_tool_list[tool.tool_id] = []
@@ -266,6 +266,8 @@ class IPUpdateHandler(webapp.RequestHandler):
                         namespace=constants.MEMCACHE_NAMESPACE_TOOLS):
                     logging.error(
                         'Failed to update sliver IP addresses in memcache.')
+                for one_tool in sliver_tool_list[tool_id]:
+                    self.put_sliver_tool(one_tool)
         return
 
     def set_sliver_tool_ips(self, sliver_tool, ipv4, ipv6):
@@ -285,9 +287,6 @@ class IPUpdateHandler(webapp.RequestHandler):
         # Update datastore
         try:
             sliver_tool.put()
-            logging.info('Succeeded to write IPs of %s (%s, %s) in datastore.',
-                         sliver_tool.fqdn, sliver_tool.sliver_ipv4,
-                         sliver_tool.sliver_ipv6)
         except db.TransactionFailedError:
             logging.error('Failed to write IPs of %s (%s, %s) in datastore.',
                           sliver_tool.fqdn, sliver_tool.sliver_ipv4,
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -246,6 +246,7 @@ class IPUpdateHandler():
                     sliver_tool_list[tool.tool_id] = []
                 updated_sliver_tool = self.set_sliver_tool(
                     sliver_tool, ipv4, ipv6, site.roundrobin)
+
                 # Update datastore if the SliverTool got updated.
                 if updated_sliver_tool:
                     self.put_sliver_tool(updated_sliver_tool)
@@ -254,7 +255,7 @@ class IPUpdateHandler():
                     sliver_tool_list[tool.tool_id].append(sliver_tool)
 
         # Update memcache.  Never set the memcache to an empty list since it's
-        # more likely that this is a Nagios failure.
+        # more likely that this is a Prometheus failure.
         if sliver_tool_list:
             for tool_id in sliver_tool_list.keys():
                 if not memcache.set(
@@ -262,7 +263,7 @@ class IPUpdateHandler():
                         sliver_tool_list[tool_id],
                         namespace=constants.MEMCACHE_NAMESPACE_TOOLS):
                     logging.error(
-                        'Failed to update sliver IP addresses in memcache.')
+                        'IPUpdateHandler: Failed to update tool %s in memcache.', tool_id)
         return
 
     def set_sliver_tool(self, sliver_tool, ipv4, ipv6, rr):
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -256,6 +256,16 @@ class IPUpdateHandler(webapp.RequestHandler):
                     sliver_tool_list[tool.tool_id] = []
                 sliver_tool_list[tool.tool_id].append(sliver_tool)
 
+        # Update memcache.  Never set the memcache to an empty list since it's
+        # more likely that this is a Nagios failure.
+        if sliver_tool_list:
+            for tool_id in sliver_tool_list.keys():
+                if not memcache.set(
+                        tool_id,
+                        sliver_tool_list[tool_id],
+                        namespace=constants.MEMCACHE_NAMESPACE_TOOLS):
+                    logging.error(
+                        'Failed to update sliver IP addresses in memcache.')
         return
 
     def set_sliver_tool_ips(self, sliver_tool, ipv4, ipv6):
@@ -272,12 +282,7 @@ class IPUpdateHandler(webapp.RequestHandler):
         return sliver_tool
 
     def put_sliver_tool(self, sliver_tool):
-        # Update memcache AND datastore here.
-        if not memcache.set(sliver_tool.tool_id,
-                            sliver_tool,
-                            namespace=constants.MEMCACHE_NAMESPACE_TOOLS):
-            logging.error('Failed to update sliver IP addresses in memcache.')
-
+        # Update datastore
         try:
             sliver_tool.put()
             logging.info('Succeeded to write IPs of %s (%s, %s) in datastore.',
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -156,33 +156,7 @@ class SiteRegistrationHandler(webapp.RequestHandler):
             logging.error('Failed to write site %s to datastore.', site.site_id)
             return False
         logging.info('Succeeded to write site %s to db', site.site_id)
-
-        tools = model.Tool.all()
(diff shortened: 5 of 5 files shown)


nkinkade:
@yachang: I've added a fairly detailed comment below, half of which I've carried over verbatim from PR #135. 

I'm also pasting below another concern I raised in PR #135, which is still something of a concern to me.

> Another concern I have with this is that the combined run time of SiteRegistrationHandler() and IPUpdateHandler() is currently going to be around 10 minutes. Meanwhile the check_status cron job will have run about 10 times, reading from memcache and unconditionally overwriting it once every minute. I worry that it will likely overwrite any updates to memcache by this new long-running cron job. What are your thoughts on this?

---

Reviewed 1 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, all discussions resolved.

---

*[server/mlabns/handlers/update.py, line 284 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/138#-LC4t2tAAxju60RmcXC2:-LC4t2tAAxju60RmcXC3:b4z6yr1) ([raw file](https://github.com/m-lab/mlab-ns/blob/af3e22a888f93b70e4bd1d014d78a1d8f1cbd114/server/mlabns/handlers/update.py#L284)):*
I've copied and pasted my comment from PR #135 below. My concern remains the same here. I'm worried that what is happening here is that this is writing a **single** SliverTool's data to the memcache key for an entire tool e.g., "ndt", or "neubot". In this case, this is going to happen every single time `put_sliver_tool` gets called, which will be hundreds of times for every tool. After this is done, it seems that, for example, the memcache key "ndt" will contain only information for a single SliverTool, instead of every single SliverTool for the tool, which is what the rest of the code is expecting.

I believe this may be the causing an issue we are seeing with mlab-ns, where every night around 1:00UTC (when the `check_ip` cron job gets run), global NDT traffic drops from a few thousand tests per minute to just hundreds.  I suspect that memcache is being reduced to information about few particular SliverTools, instead of all of them.

See the code links in the pasted message below for examples, but (for better or worse) I believe that memcache looks like this (in pseudo code):

```
memcache = {
    'ndt': <python object containing ALL ndt SliverTools>,
    'ndt_ssl':  <python object containing ALL ndt_ssl SliverTools>,
    'mobiperf':  <python object containing ALL mobiperf SliverTools>,
    etc.,
    etc.,
}
```

This is, of course, very different than the way the datastore looks.

I believe that the solution here is to not write to memcache here in `put_sliver_tool`, and it should only write to the datastore, which does have individual records keyed for every single SliverTool.  The SliverTools [already get written to memcache a few lines up in the main `get()` function](https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/handlers/update.py#L277).

> Will this overwrite the same memcache key (tool_id, e.g., ndt, neubot, etc.) three different times with different data? tool_id comes from an outer loop, in which tool.tool_id will not change in the inner loop. The inner loop iterates over ['mlab1', 'mlab2', 'mlab3'].
>
> The other possible issue I see here is that this is a very large departure from my understanding of how memcache is currently used. This addition seems to be storing a single sliver_tool in memcache, keyed on tool_id. However, in other places the memcache key tool_id stores a large python object (a dict or list of dicts) and the code expects, for example, that something like memcache.get('ndt') is going to return every single NDT sliver tool, not just one single sliver tool. Examples:

> https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/db/sliver_tool_fetcher.py#L200
> https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/handlers/update.py#L509
> https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/handlers/update.py#L279

> I believe this is why the memcache viewer in the GCP console doesn't return anything readable, because there are only a small handful of memcache keys, one for each tool (ndt, ndt_ssl, neubot, mobiperf, npad, ooni), and behind that key is every single sliver_tool for that given tool.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/138#-:-LC4wMgVBoqzN1dCJy0q:bk5rxg8)*

yachang:
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[server/mlabns/handlers/update.py, line 284 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/138#-LC4t2tAAxju60RmcXC2:-LC5AUbW4tAe5OCwtIB0:b-xfb4p) ([raw file](https://github.com/m-lab/mlab-ns/blob/af3e22a888f93b70e4bd1d014d78a1d8f1cbd114/server/mlabns/handlers/update.py#L284)):*

I've copied and pasted my comment from PR #135 below. My concern remains the same here. I'm worried that what is happening here is that this is writing a **single** SliverTool's data to the memcache key for an entire tool e.g., "ndt", or "neubot". In this case, this is going to happen every single time `put_sliver_tool` gets called, which will be hundreds of times for every tool. After this is done, it seems that, for example, the memcache key "ndt" will contain only information for a single SliverTool, instead of every single SliverTool for the tool, which is what the rest of the code is expecting.

I believe this may be the causing an issue we are seeing with mlab-ns, where every night around 1:00UTC (when the `check_ip` cron job gets run), global NDT traffic drops from a few thousand tests per minute to just hundreds.  I suspect that memcache is being reduced to information about few particular SliverTools, instead of all of them.

See the code links in the pasted message below for examples, but (for better or worse) I believe that memcache looks like this (in pseudo code):

```
memcache = {
    'ndt': <python object containing ALL ndt SliverTools>,
    'ndt_ssl':  <python object containing ALL ndt_ssl SliverTools>,
    'mobiperf':  <python object containing ALL mobiperf SliverTools>,
    etc.,
    etc.,
}
```

This is, of course, very different than the way the datastore looks.

I believe that the solution here is to not write to memcache here in `put_sliver_tool`, and it should only write to the datastore, which does have individual records keyed for every single SliverTool.  The SliverTools [already get written to memcache a few lines up in the main `get()` function](https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/handlers/update.py#L277).

> Will this overwrite the same memcache key (tool_id, e.g., ndt, neubot, etc.) three different times with different data? tool_id comes from an outer loop, in which tool.tool_id will not change in the inner loop. The inner loop iterates over ['mlab1', 'mlab2', 'mlab3'].
>
> The other possible issue I see here is that this is a very large departure from my understanding of how memcache is currently used. This addition seems to be storing a single sliver_tool in memcache, keyed on tool_id. However, in other places the memcache key tool_id stores a large python object (a dict or list of dicts) and the code expects, for example, that something like memcache.get('ndt') is going to return every single NDT sliver tool, not just one single sliver tool. Examples:

> https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/db/sliver_tool_fetcher.py#L200
> https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/handlers/update.py#L509
> https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/handlers/update.py#L279

> I believe this is why the memcache viewer in the GCP console doesn't return anything readable, because there are only a small handful of memcache keys, one for each tool (ndt, ndt_ssl, neubot, mobiperf, npad, ooni), and behind that key is every single sliver_tool for that given tool.
</blockquote></details>

Thanks a lot for the detailed review!

Thanks for pointing out that we set memcache twice in check_ip which might be the reason it took 10 minutes now. I remove the dup operation and make sure


1. memcache are written as tool_id -> list of SliverTool
2. DataStore are updated only for modified SliverTool
3. DataStore should syn with memcache for whatever change, which make the mutex operation unnecessary.

It now takes about <5 minutes to complete the whole job (check_site + check_ip)

Once it is checked in, we can remove the check_ip job (in a separate PR)

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/138)*

yachang:
I deploy it on mlab-nsteating, and turn on dub01, which is working : https://pantheon.corp.google.com/datastore/entities/query?project=mlab-nstesting&folder&organizationId=433637338589&ns=&kind=SliverTool&filter=7%2Fsite_id%7CSTR%7CEQ%7C5%2Fdub01

---

Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/138#-:-LC7IKpv0m2fsg0TVHqB:b-8w0047)*

nkinkade:
@yachang: Thank you very much for this. I have a couple of very small, trivial comments, but otherwise this looks great!

---

Reviewed 2 of 4 files at r2.
Review status: 3 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[server/cron.yaml, line 10 at r2](https://reviewable.io/reviews/m-lab/mlab-ns/138#-LCAiUrI5ek2nUC03lCr:-LCAiUrI5ek2nUC03lCs:b-xkkgrt) ([raw file](https://github.com/m-lab/mlab-ns/blob/ce773ab4e49b9ad165e8f4ce7204787b268a2f1f/server/cron.yaml#L10)):*
> ```YAML
>   schedule: every 1 minutes
> 
> # Check for new sites every 24 hours, starting at 00:00
> ```

Since `check_ip` will soon be gone, could you update this comment to indicate that not only will this cron job be checking for new sites, but will also be checking for new and/or updated IP addresses?

---

*[server/cron.yaml, line 13 at r2](https://reviewable.io/reviews/m-lab/mlab-ns/138#-LCAicK49m_qi4eWYwtm:-LCAicK49m_qi4eWYwtn:b-5zuwck) ([raw file](https://github.com/m-lab/mlab-ns/blob/ce773ab4e49b9ad165e8f4ce7204787b268a2f1f/server/cron.yaml#L13)):*
> ```YAML
> - description: Check sites
>   url: /cron/check_site
>   schedule: every day 00:00
> ```

This is perhaps outside the scope of your current PR, but one 00:00 UTC is right around prime time Internet usage in the United States. Because of this, any bugs in the code that we don't know about could possibly be affecting a lot more people than it might otherwise. Eventually this will be less relevant once we have larger deployments in Asia and Africa, but for now the vast bulk of our traffic is from the U.S. and Europe. Maybe we could change this cron job to fire at a time of day that is in the middle of the night or early morning for the U.S. and Europe instead of UTC 00:00? Maybe something like 6:00 UTC?

---

*[server/mlabns/handlers/update.py, line 284 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/138#-LC4t2tAAxju60RmcXC2:-LCAjmuD4CClDgITlG7x:b5cy6if) ([raw file](https://github.com/m-lab/mlab-ns/blob/af3e22a888f93b70e4bd1d014d78a1d8f1cbd114/server/mlabns/handlers/update.py#L284)):*
> 3. DataStore should syn with memcache for whatever change, which make the mutex operation unnecessary.

I'm still not sure that `check_status` won't interfere with this job. This job will now sync the datastore and memcache well, but what will stop `check_status` from overwriting memcache with old/bad information? That said, this issue has always existed and is not necessarily within the scope of this PR. Indeed, I think we should probably fix the current bug in mlab-ns (which this PR does), then address the race condition between the cron jobs in a separate PR. What do you think?

---

*[server/mlabns/handlers/update.py, line 249 at r2](https://reviewable.io/reviews/m-lab/mlab-ns/138#-LCAg5p6-1JIIqbyBc6v:-LCAg5p7CZrbiDgyIwVx:bz0mlqd) ([raw file](https://github.com/m-lab/mlab-ns/blob/ce773ab4e49b9ad165e8f4ce7204787b268a2f1f/server/mlabns/handlers/update.py#L249)):*
> ```Python
>                             tool.tool_id, fqdn))
>                     continue
>                 if tool.tool_id not in sliver_tool_list:
> ```

Tiny nit: could you add a newline above this so that it is visually clear that it is not part of the `if` block above it? Same for statements below?

---

*[server/mlabns/handlers/update.py, line 261 at r2](https://reviewable.io/reviews/m-lab/mlab-ns/138#-LCAgw0x8MRnhXcSvc1P:-LCAgw0x8MRnhXcSvc1Q:bkn2n50) ([raw file](https://github.com/m-lab/mlab-ns/blob/ce773ab4e49b9ad165e8f4ce7204787b268a2f1f/server/mlabns/handlers/update.py#L261)):*
> ```Python
> 
>         # Update memcache.  Never set the memcache to an empty list since it's
>         # more likely that this is a Nagios failure.
> ```

This is outside the scope of the work you are doing here, but since you have this open PR, would you be willing to replace `Nagios` with `Prometheus` in this comment?

---

*[server/mlabns/handlers/update.py, line 269 at r2](https://reviewable.io/reviews/m-lab/mlab-ns/138#-LCAh9X30WlWI1ZS6cti:-LCAh9X30WlWI1ZS6ctj:bbk72hi) ([raw file](https://github.com/m-lab/mlab-ns/blob/ce773ab4e49b9ad165e8f4ce7204787b268a2f1f/server/mlabns/handlers/update.py#L269)):*
> ```Python
>                         namespace=constants.MEMCACHE_NAMESPACE_TOOLS):
>                     logging.error(
>                         'Failed to update sliver IP addresses in memcache.')
> ```

The above log message isn't accurate any longer. Maybe it should read something more like the following?

`logging.error('IPUpdateHandler: Failed to update tool %s in memcache.', tool_id)`

---

*[server/mlabns/handlers/update.py, line 305 at r2](https://reviewable.io/reviews/m-lab/mlab-ns/138#-LCAhzTm85fBlVpqfBea:-LCAhzTm85fBlVpqfBeb:b6k39q6) ([raw file](https://github.com/m-lab/mlab-ns/blob/fc17d3bbc5b207a1b5db9e0ac4224209bb4c7593/server/mlabns/handlers/update.py#L305)):*
> ```Python
>             logging.info('Succeeded to write IPs of %s (%s, %s) in datastore.',
>                          sliver_tool.fqdn, sliver_tool.sliver_ipv4,
>                          sliver_tool.sliver_ipv6)
> ```

Thank you for removing this logging which wasn't helping anyone, but only serving to make the logs difficult to read for more important things!

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/138#-:-LCAiF_L2JeAV_6VSmay:b-dex5s3)*

yachang:
Review status: 3 of 5 files reviewed at latest revision, 7 unresolved discussions.

---

*[server/mlabns/handlers/update.py, line 284 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/138#-LC4t2tAAxju60RmcXC2:-LCEadW5-ZZtKvlFMUZa:b-m9y5pe) ([raw file](https://github.com/m-lab/mlab-ns/blob/af3e22a888f93b70e4bd1d014d78a1d8f1cbd114/server/mlabns/handlers/update.py#L284)):*

> 3. DataStore should syn with memcache for whatever change, which make the mutex operation unnecessary.

I'm still not sure that `check_status` won't interfere with this job. This job will now sync the datastore and memcache well, but what will stop `check_status` from overwriting memcache with old/bad information? That said, this issue has always existed and is not necessarily within the scope of this PR. Indeed, I think we should probably fix the current bug in mlab-ns (which this PR does), then address the race condition between the cron jobs in a separate PR. What do you think?
</blockquote></details>

The biggest bug fixed in this PR is:

line 253

sliver_tool_list[tool.tool_id].append(updated_sliver_tool)

So the memcache got the updated info as well and will be updated after DataStore got updated.

Ideally, we can delete the DataStore update code (line 252), and trust memcache will do the same job in a minute. But based on review from last PR, we both agree to keep the code for DataStore for safety.

---

*[server/mlabns/handlers/update.py, line 249 at r2](https://reviewable.io/reviews/m-lab/mlab-ns/138#-LCAg5p6-1JIIqbyBc6v:-LCEaMLV9POscRcx0u4J:b-896fix) ([raw file](https://github.com/m-lab/mlab-ns/blob/ce773ab4e49b9ad165e8f4ce7204787b268a2f1f/server/mlabns/handlers/update.py#L249)):*

Tiny nit: could you add a newline above this so that it is visually clear that it is not part of the `if` block above it? Same for statements below?
</blockquote></details>

Done.

---

*[server/mlabns/handlers/update.py, line 261 at r2](https://reviewable.io/reviews/m-lab/mlab-ns/138#-LCAgw0x8MRnhXcSvc1P:-LCEaSJ71G4IFwQS_0a7:b-896fix) ([raw file](https://github.com/m-lab/mlab-ns/blob/ce773ab4e49b9ad165e8f4ce7204787b268a2f1f/server/mlabns/handlers/update.py#L261)):*

This is outside the scope of the work you are doing here, but since you have this open PR, would you be willing to replace `Nagios` with `Prometheus` in this comment?
</blockquote></details>

Done.

---

*[server/mlabns/handlers/update.py, line 269 at r2](https://reviewable.io/reviews/m-lab/mlab-ns/138#-LCAh9X30WlWI1ZS6cti:-LCEa_LT01nif2xzfO1H:b-896fix) ([raw file](https://github.com/m-lab/mlab-ns/blob/ce773ab4e49b9ad165e8f4ce7204787b268a2f1f/server/mlabns/handlers/update.py#L269)):*

The above log message isn't accurate any longer. Maybe it should read something more like the following?

`logging.error('IPUpdateHandler: Failed to update tool %s in memcache.', tool_id)`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/138)*

yachang:
Thanks a lot! I have addressed your comments. Could you take a look?

---

Review status: 3 of 5 files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/138#-:-LCEcCKP35EA8s298pKv:bc1yhby)*

nkinkade:
:lgtm: Thanks, @yachang!

---

Reviewed 1 of 1 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[server/mlabns/handlers/update.py, line 284 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/138#-LC4t2tAAxju60RmcXC2:-LCEtW0NDmLhcuESVCiC:b-a3g0ei) ([raw file](https://github.com/m-lab/mlab-ns/blob/af3e22a888f93b70e4bd1d014d78a1d8f1cbd114/server/mlabns/handlers/update.py#L284)):*

The biggest bug fixed in this PR is:

line 253

sliver_tool_list[tool.tool_id].append(updated_sliver_tool)

So the memcache got the updated info as well and will be updated after DataStore got updated.

Ideally, we can delete the DataStore update code (line 252), and trust memcache will do the same job in a minute. But based on review from last PR, we both agree to keep the code for DataStore for safety.
</blockquote></details>

I agree. Thanks!

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/138#-:-LCEtg8g7YWX0xOeUWYC:b-4kthzs)*

PR #146, opened 2018-06-13
Preserves every sliver tool in memcache, whether it exists in monitoring data or not.

nkinkade:
Currently, if a slivertool exists in mlab-ns's memcache, but does not exist in the results returned by Prometheus, it will be implicitly removed from memcache. If there is some sort of major failure in Prometheus, or a bug in our PromQL queries, that causes some or all slivertools to not be present in the monitoring data returned by a query, memcache may only know about some slivertools. If the failure is large, memcache may get close to empty and global traffic will possibly be directed at just the few remaining slivertools.

This PR ensures that all slivertools in memcache at the time the `check_status` cron job is run will remain in memcache, even if it means that the true status of the slivertool is unknown.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/146)

### DIFF
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -443,6 +443,10 @@ class StatusUpdateHandler(webapp.RequestHandler):
                     slice_status[sliver_tool.fqdn][
                         'tool_extra'] = sliver_tool.tool_extra + '(Family "_ipv6" for sliver not known by monitoring).'
                 else:
+                    # If monitoring data doesn't exist for this tool, append
+                    # the sliver_tool unmodified to the list that gets written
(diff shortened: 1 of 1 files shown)


stephen-soltesz:
:lgtm: -- please add a todo or create an issue to test the condition that this change should guard against.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/146#-:-LEvC6m36PKQLlS3nuWP:b82u49z)*

nkinkade:
Issue created. https://github.com/m-lab/mlab-ns/issues/147.

PR #149, opened 2018-06-15
Return a record for every service, even if a particular metric is missing.

nkinkade:
In a recent incident that affected global traffic, we came to understand that we cannot count on any given metric actually existing for any particular instantaneous query. This PR modifies the Prometheus queries mlab-ns issues such that even if a metric is missing, a result (of 0) will still be returned for machine/service.

**IMPORTANT CONSIDERATION**: In PR #146 we added a feature whereby if status for a particular "tool" on a particular machine is missing from the data returned by Prometheus, we don't touch the existing status of the slivertool in mlab-ns. That is, in the absence of data, we do nothing to change the status quo. This PR is going to affect that such that missing metrics will mark a slivertool as down in mlab-ns. This may be what we want. That is, we may want to say: If we don't know the status of a slivertool, then flag it as down until we do know the true status. However, the downside to this is that we could end up in another situation where, say, the blackbox_exporter goes down for 5 or 10 minutes removing a bunch of metrics. With this PR, mlab-ns will flag everything for which there is some missing metric down. A service interruption in the Prometheus flow could take down most of the fleet in this way.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/149)

### DIFF
--- a/server/mlabns/util/prometheus_status.py
+++ b/server/mlabns/util/prometheus_status.py
@@ -41,124 +41,120 @@ QUERIES = {
     'ndt': textwrap.dedent("""\
         min by (experiment, machine) (
           (
-            (
-              (probe_success{service="ndt_raw"}) +
-              ON (experiment, machine) (script_success{service="ndt_e2e"}) +
-              ON (experiment, machine) ((vdlimit_used{experiment="ndt.iupui"} /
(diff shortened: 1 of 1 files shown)


nkinkade:
Closing this PR, which is superseded by #154.

Issue #97, opened 2017-04-28
Race condition when reading/writing to memcache


nkinkade:
There is a race condition created by two cron jobs. The [cron job `check_ip`](https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/handlers/update.py#L174) runs once an hour, and splats its updates to memcache. The job takes around 2 or 3 minutes to complete. The [cron job `check_status` runs every minute](https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/handlers/update.py#L329) and also splats its results to memcache. These two jobs are almost surely stepping on each other in some way in many cases, most likely the latter on the former.

The [GAE docs recommend using gets() and cas()](https://cloud.google.com/appengine/docs/standard/python/memcache/using) to help get around race conditions, but with the current architecture of mlab-ns there is no apparently (to me) easy way to implement this without some significant changes.

nkinkade:
I believe this will be solved by PR #150. Once PR #150 is merge, close this issue.

nkinkade:
PR #150 was merged a long time ago. Closing.

PR #150, opened 2018-06-19
Use the datastore as the source of truth, not memcache.

nkinkade:
The primary purpose of this PR (and the smallest amount of changes) is to change the `check_status` cron job to use the datastore as the source of truth instead of what happens to be in memcache. Currently, `check_status` reads what is in memcache for a given tool and then iterates over it to determine if there are any status changes. The problem with this is that memcache (as we use it) is a "best-effort" service, and the contents of memcache might disappear at any moment for any reason.

From: https://cloud.google.com/appengine/articles/best-practices-for-app-engine-memcache

> Data stored in Memcache can be evicted at any time, so applications should be structured in a way that does not depend on the presence of entries. [...] Shared Memcache is a no-cost, best-effort service, and data stored with this service can be evicted at any time.

Furthermore, if data gets evicted from memcache, or a bug in our code removes it accidentally, as was the case in a recent incident, `check_status` does not try to verify that what is in memcache represents the full platform inventory. Thus if memcache is somehow truncated, mlab-ns will only updates status for and advertise to the world this truncated set of slivertools.

This PR causes `check_status` to read all known slivertools from the **datastore** instead of memcache. When it is done updating status on these slivertools, it first writes the updated data back to the **datastore**. When everything is done, it dumps this updated status to memcache.

This PR also stops the `IPUpdateHandler()` from writing anything to memcache, which should 
 resolve issue #97, since now only a single job is ever writing anything to memcache. 

With these changes, if memcache get corrupted or truncated for any reason, it will be updated with the full set of slivertools in the datastore the next time `check_status` runs, which will be less than 1m.

The second goal of this PR, and the one which actually makes up the majority of the code changes, is to improve the performance of the `IPUpdateHandler()` code. Currently this code takes between 3-5 minutes to run. The reason for this long run time is that it currently make 15,000 to 20,000 individual datastore queries. This PR fixes that by fetching all the data from the datastore that it needs up front, in 3 big queries. With these changes the runtime for `IPUpdateHandler()` is only about 25s.

One other small fix in the PR is that currently [there is a constant in util/constants.py](https://github.com/m-lab/mlab-ns/blob/master/server/mlabns/util/constants.py#L30) that configures mlab-ns to never fetch more than 500 records from the datastore. We now have more than 500 nodes, and this has been excluding some slivertools from getting updated.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/150)

### DIFF
--- a/server/mlabns/db/sliver_tool_fetcher.py
+++ b/server/mlabns/db/sliver_tool_fetcher.py
@@ -255,5 +255,6 @@ class SliverToolFetcherDatastore(object):
         if not tool_properties.all_slivers:
             results = _filter_choose_one_host_per_site(results)
 
-        logging.info('%d sliver tools found in Datastore.', len(results))
+        logging.info('{}: {} sliver tools found in datastore.'.format(
+            tool_properties.tool_id, len(results)))
         return results
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -429,7 +429,7 @@ class StatusUpdateHandler(webapp.RequestHandler):
                         'status'] = message.STATUS_OFFLINE
                     # Don't wipe out the IPv4 'tool_extra', just append to it.
                     slice_status[sliver_tool.fqdn][
-                        'tool_extra'] = sliver_tool.tool_extra + '(Family "_ipv6" for sliver not known by monitoring).'
+                        'tool_extra'] = constants.PROMETHEUS_TOOL_EXTRA + '(Family "_ipv6" for sliver not known by monitoring).'
                 else:
                     # If monitoring data doesn't exist for this tool, append
                     # the sliver_tool unmodified to the list, since in the
--- a/server/mlabns/handlers/update.py
+++ b/server/mlabns/handlers/update.py
@@ -181,7 +181,11 @@ class IPUpdateHandler():
             logging.error('Cannot open %s.', self.IP_LIST_URL)
             return util.send_not_found(self)
 
-        sliver_tool_list = {}
+        # Fetch all data that we are going to need from the datastore up front.
+        sites = list(model.Site.all().fetch())
+        tools = list(model.Tool.all().fetch())
+        slivertools = list(model.SliverTool.all().fetch())
+
         for line in lines:
             # Expected format: "FQDN,IPv4,IPv6" (IPv6 can be an empty string).
             line_fields = line.split(',')
@@ -206,66 +210,50 @@ class IPUpdateHandler():
                 continue
 
             # If mlab-ns does not support this site, then skip it.
-            site = model.Site.gql('WHERE site_id=:site_id',
-                                  site_id=site_id).get()
-            if site == None:
+            site = list(filter(lambda s: s['site_id'] == site_id, sites))
+            if len(site) == 0:
                 logging.info('mlab-ns does not support site %s.', site_id)
                 continue
+            else:
+                site = site[0]
 
             # If mlab-ns does not serve/support this slice, then skip it. Note:
             # a given slice_id might have multiple tools (e.g., iupui_ndt has
             # both 'ndt' and 'ndt_ssl' tools.
-            tools = model.Tool.gql('WHERE slice_id=:slice_id',
-                                   slice_id=slice_id)
-            if tools.count() == 0:
+            tools = list(filter(lambda t: t['slice_id'] == slice_id, tools))
+
+            if len(tools) == 0:
                 continue
 
-            for tool in tools.run():
-                # Query the datastore to see if this sliver_tool exists there.
-                sliver_tool_gql = model.SliverTool.gql(
-                    'WHERE fqdn=:fqdn AND tool_id=:tool_id',
-                    fqdn=fqdn,
-                    tool_id=tool.tool_id)
+            for tool in tools:
+                # See if this sliver_tool already exists in the datastore.
+                slivertool = list(filter(
+                    lambda st: st['fqdn'] == fqdn and st['tool_id'] == tool['tool_id'],
+                    slivertools))
 
                 # Check to see if the sliver_tool already exists in the
                 # datastore. If not, add it to the datastore.
-                if sliver_tool_gql.count() == 1:
-                    sliver_tool = sliver_tool_gql.get(
-                        batch_size=constants.GQL_BATCH_SIZE)
-                elif sliver_tool_gql.count() == 0:
+                if len(slivertool) == 1:
+                    sliver_tool = slivertool[0]
+                elif len(slivertool) == 0:
                     logging.info(
                         'For tool %s, fqdn %s is not in datastore.  Adding it.',
-                        tool.tool_id, fqdn)
+                        tool['tool_id'], fqdn)
                     sliver_tool = self.initialize_sliver_tool(tool, site,
                                                               server_id, fqdn)
                 else:
                     logging.error(
                         'Error, or too many sliver_tools returned for {}:{}.'.format(
-                            tool.tool_id, fqdn))
+                            tool['tool_id'], fqdn))
                     continue
-                if tool.tool_id not in sliver_tool_list:
-                    sliver_tool_list[tool.tool_id] = []
+
                 updated_sliver_tool = self.set_sliver_tool(
-                    sliver_tool, ipv4, ipv6, site.roundrobin)
+                    sliver_tool, ipv4, ipv6, site['roundrobin'])
 
                 # Update datastore if the SliverTool got updated.
                 if updated_sliver_tool:
                     self.put_sliver_tool(updated_sliver_tool)
-                    sliver_tool_list[tool.tool_id].append(updated_sliver_tool)
-                else:
-                    sliver_tool_list[tool.tool_id].append(sliver_tool)
 
-        # Update memcache.  Never set the memcache to an empty list since it's
-        # more likely that this is a Prometheus failure.
-        if sliver_tool_list:
-            for tool_id in sliver_tool_list.keys():
-                if not memcache.set(
-                        tool_id,
-                        sliver_tool_list[tool_id],
-                        namespace=constants.MEMCACHE_NAMESPACE_TOOLS):
-                    logging.error(
-                        'IPUpdateHandler: Failed to update tool %s in memcache.',
-                        tool_id)
         return
 
     def set_sliver_tool(self, sliver_tool, ipv4, ipv6, rr):
@@ -275,14 +263,14 @@ class IPUpdateHandler():
         if not ipv4:
             ipv6 = message.NO_IP_ADDRESS
 
-        if not sliver_tool.sliver_ipv4 == ipv4:
-            sliver_tool.sliver_ipv4 = ipv4
+        if not sliver_tool['sliver_ipv4'] == ipv4:
+            sliver_tool['sliver_ipv4'] = ipv4
             updated = True
-        if not sliver_tool.sliver_ipv6 == ipv6:
-            sliver_tool.sliver_ipv6 = ipv6
+        if not sliver_tool['sliver_ipv6'] == ipv6:
+            sliver_tool['sliver_ipv6'] = ipv6
             updated = True
-        if not sliver_tool.roundrobin == rr:
-            sliver_tool.roundrobin = rr
+        if not sliver_tool['roundrobin'] == rr:
+            sliver_tool['roundrobin'] = rr
             updated = True
 
         if updated:
@@ -295,21 +283,21 @@ class IPUpdateHandler():
             sliver_tool.put()
         except db.TransactionFailedError:
             logging.error('Failed to write IPs of %s (%s, %s) in datastore.',
-                          sliver_tool.fqdn, sliver_tool.sliver_ipv4,
-                          sliver_tool.sliver_ipv6)
+                          sliver_tool['fqdn'], sliver_tool['sliver_ipv4'],
+                          sliver_tool['sliver_ipv6'])
 
     def initialize_sliver_tool(self, tool, site, server_id, fqdn):
-        sliver_tool_id = model.get_sliver_tool_id(tool.tool_id, tool.slice_id,
-                                                  server_id, site.site_id)
+        sliver_tool_id = model.get_sliver_tool_id(
+            tool['tool_id'], tool['slice_id'], server_id, site['site_id'])
 
         return model.SliverTool(
-            tool_id=tool.tool_id,
-            slice_id=tool.slice_id,
-            site_id=site.site_id,
+            tool_id=tool['tool_id'],
+            slice_id=tool['slice_id'],
+            site_id=site['site_id'],
             server_id=server_id,
             fqdn=fqdn,
-            server_port=tool.server_port,
-            http_port=tool.http_port,
+            server_port=tool['server_port'],
+            http_port=tool['http_port'],
             # IP addresses will be updated by the IPUpdateHandler.
             sliver_ipv4=message.NO_IP_ADDRESS,
             sliver_ipv6=message.NO_IP_ADDRESS,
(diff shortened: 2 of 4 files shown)


nkinkade:
Oh, one thing I forgot add to the description was that this PR also fulfills a TODO in update.py to implement a `TESTING_IP_LIST_URL` which points to mlab-host-ips.txt in the mlab-sandbox project.

---

Review status: 0 of 1 LGTMs obtained

---



*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/150#-:-LFPxtGw8EcTfH8CXf_k:b-av47xd)*

yachang:
There are two major concerns for this logic change:

1. check_status running once every minutes. I highly doubt that read/write from datastore for all 500+ sliver_tools can be done within one minute. The initial motivation for using memcache here is to speed up the check_status job.

2. If there is no performance issue and check_status always read/write from memcache, then there is no reason to keep memcache at all. Mlab-ns code can be simplified even further.

---

Review status: 0 of 1 LGTMs obtained

---



*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/150#-:-LFSZ92nBvKNpiqg1f7z:b-j5hy1w)*

nkinkade:
@yachang: I'll address your concerns in the same order you placed them:

1. We don't have to wonder! This new code, is running in mlab-nstesting. Looking at the logs, the `check_status` job only takes about 20 seconds to run (measuring the time between the first and last log message). This is about the same amount of time the one using memcache has always been taking. Indeed it may actually be less, because I recall the current one taking up to 40 seconds. Also, remember that this code isn't writing each of 500 records to the datastore individually, but is `put()`ing them all at once. Presumably the Google `db` code optimizes for this.  So, using real world observations, do you feel that this job isn't completing fast enough?

2. We still need memcache to serve up client requests! Those requests are very time/latency sensitive, but updating the datastore isn't as much.

---

Review status: 0 of 1 LGTMs obtained

---



*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/150#-:-LFSdJMBDFjERVDw52QQ:b7il9dn)*

yachang:
Review status: 0 of 1 LGTMs obtained

---

*[server/mlabns/handlers/update.py, line 268 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/150#-LFThA6T9zkCTZwRo0_k:-LFThA6T9zkCTZwRo0_l:b-peq1o0) ([raw file](https://github.com/m-lab/mlab-ns/blob/ec3d8107cf8c7d6591018ed37869f33c9e19f991/server/mlabns/handlers/update.py#L268)):*
> ```Python
>                     logging.error(
>                         'IPUpdateHandler: Failed to update tool %s in memcache.',
>                         tool_id)
> ```

It will cause problem to delete this memcache sync here. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/150)*

yachang:
Review status: 0 of 1 LGTMs obtained

---

*[server/mlabns/db/sliver_tool_fetcher.py, line 161 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/150#-LFU1MaT7E_KL62ed6g8:-LFU1MaT7E_KL62ed6g9:b-ywr1cv) ([raw file](https://github.com/m-lab/mlab-ns/blob/1c6ab4a427ddb9d9319f4104b18fc3f2e984cd20/server/mlabns/db/sliver_tool_fetcher.py#L161)):*
> ```Python
>             A list of SliverTool objects that match the specified criteria.
>         """
>         results = self._memcache_fetcher.fetch(tool_properties)
> ```

By what you described, the logic here needs change:

This logic in this func is to fetch memcache first, if not existing, try to fetch datastore.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/150)*

nkinkade:
This isn't something that we've done on prior releases. Is artificial traffic that you generate yourself (by making calls directly through your browser) not enough to be sure that requests are getting fast replies. Also, don't forget, requests from clients are still getting served from data stored in memcache. Nothing has changed about that process. The change in the PR is that the cron job reads from the datastore instead of memcache.

---

Review status: 0 of 1 LGTMs obtained

---

*[server/mlabns/db/sliver_tool_fetcher.py, line 161 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/150#-LFU1MaT7E_KL62ed6g8:-LFU4jnGCUvMkg9kfwGX:b-tkf4tx) ([raw file](https://github.com/m-lab/mlab-ns/blob/1c6ab4a427ddb9d9319f4104b18fc3f2e984cd20/server/mlabns/db/sliver_tool_fetcher.py#L161)):*

By what you described, the logic here needs change:

This logic in this func is to fetch memcache first, if not existing, try to fetch datastore.
</blockquote></details>

This function doesn't get called by `check_status` anymore. It now calls `SliverToolFetcherDatastore` directly. Refer to this line of the PR:

https://github.com/m-lab/mlab-ns/pull/150/files#diff-bdb66b35c5806c4f5ebf7962adf4d831R424

---

*[server/mlabns/handlers/update.py, line 168 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/150#-LFTeO-Z8u79fLV-Dcjf:-LFU5AI07E3k9-otucil:b-896fix) ([raw file](https://github.com/m-lab/mlab-ns/blob/1c6ab4a427ddb9d9319f4104b18fc3f2e984cd20/server/mlabns/handlers/update.py#L168)):*

Thanks for adding this!
</blockquote></details>

Done.

---

*[server/mlabns/handlers/update.py, line 268 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/150#-LFThA6T9zkCTZwRo0_k:-LFU5c8H8iErDVqIqhy3:b-99vald) ([raw file](https://github.com/m-lab/mlab-ns/blob/ec3d8107cf8c7d6591018ed37869f33c9e19f991/server/mlabns/handlers/update.py#L268)):*

It will cause problem to delete this memcache sync here. 
</blockquote></details>

Would you explain your concern in more detail?

The way I'm thinking about it, it shouldn't be a problem if this doesn't write to memcache, since the next time `check_status` runs it will read any updated data written to the datastore by `IPUpdateHandler`. Since `check_status` runs every minute, then the delay for new data written to the datastore by `IPUpdateHandler` making it into memcache will be less than 60 seconds.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/150#-:-LFU6LUsEm3kjaXhK5fs:bwjq8zd)*

yachang:
Review status: 0 of 1 LGTMs obtained

---

*[server/mlabns/db/sliver_tool_fetcher.py, line 161 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/150#-LFU1MaT7E_KL62ed6g8:-LFUAhoZDDZzcgf0Sk6t:bd6dkiz) ([raw file](https://github.com/m-lab/mlab-ns/blob/1c6ab4a427ddb9d9319f4104b18fc3f2e984cd20/server/mlabns/db/sliver_tool_fetcher.py#L161)):*

This function doesn't get called by `check_status` anymore. It now calls `SliverToolFetcherDatastore` directly. Refer to this line of the PR:

https://github.com/m-lab/mlab-ns/pull/150/files#diff-bdb66b35c5806c4f5ebf7962adf4d831R424
</blockquote></details>

Is this func called anywhere else? If not, we can clean up the code here (not necessarily this PR)

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/150)*

yachang:
Review status: 0 of 1 LGTMs obtained

---

*[server/mlabns/handlers/update.py, line 268 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/150#-LFThA6T9zkCTZwRo0_k:-LFUGFD1EttekepIY9KP:b-hkf0qv) ([raw file](https://github.com/m-lab/mlab-ns/blob/ec3d8107cf8c7d6591018ed37869f33c9e19f991/server/mlabns/handlers/update.py#L268)):*

Would you explain your concern in more detail?

The way I'm thinking about it, it shouldn't be a problem if this doesn't write to memcache, since the next time `check_status` runs it will read any updated data written to the datastore by `IPUpdateHandler`. Since `check_status` runs every minute, then the delay for new data written to the datastore by `IPUpdateHandler` making it into memcache will be less than 60 seconds.
</blockquote></details>

 IPUpdateHandler run in daily base, as part of cron job check_site, right?

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/150)*

nkinkade:
Review status: 0 of 1 LGTMs obtained

---

*[server/mlabns/db/sliver_tool_fetcher.py, line 161 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/150#-LFU1MaT7E_KL62ed6g8:-LFUHG1F09KRFQmvxfri:b-tivi9r) ([raw file](https://github.com/m-lab/mlab-ns/blob/1c6ab4a427ddb9d9319f4104b18fc3f2e984cd20/server/mlabns/db/sliver_tool_fetcher.py#L161)):*

Is this func called anywhere else? If not, we can clean up the code here (not necessarily this PR)
</blockquote></details>

Yes, `SliverToolFetcher()` is definitely still called elsewhere, just not the `check_status` job.

---

*[server/mlabns/handlers/update.py, line 268 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/150#-LFThA6T9zkCTZwRo0_k:-LFUHTJj2XVgWpdMpIpM:b-4jjfh7) ([raw file](https://github.com/m-lab/mlab-ns/blob/ec3d8107cf8c7d6591018ed37869f33c9e19f991/server/mlabns/handlers/update.py#L268)):*

 IPUpdateHandler run in daily base, as part of cron job check_site, right?
</blockquote></details>

Yes, `IPUpdateHandler` runs daily (via `check_site`) once a day at 06:00 UTC.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/150)*

yachang:
Review status: 0 of 1 LGTMs obtained

---

*[server/mlabns/handlers/update.py, line 268 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/150#-LFThA6T9zkCTZwRo0_k:-LFUOSMy9chx_yaxo-08:bg1hcyo) ([raw file](https://github.com/m-lab/mlab-ns/blob/ec3d8107cf8c7d6591018ed37869f33c9e19f991/server/mlabns/handlers/update.py#L268)):*

Yes, `IPUpdateHandler` runs daily (via `check_site`) once a day at 06:00 UTC.
</blockquote></details>

So you just said "the delay for new data written to the datastore by IPUpdateHandler making it into memcache will be less than 60 seconds"

it is not right.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/150)*

nkinkade:
Review status: 0 of 1 LGTMs obtained

---

*[server/mlabns/handlers/update.py, line 268 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/150#-LFThA6T9zkCTZwRo0_k:-LFUc9AQ6tM9PU9ywQX0:bdg5pt0) ([raw file](https://github.com/m-lab/mlab-ns/blob/ec3d8107cf8c7d6591018ed37869f33c9e19f991/server/mlabns/handlers/update.py#L268)):*

So you just said "the delay for new data written to the datastore by IPUpdateHandler making it into memcache will be less than 60 seconds"

it is not right.
</blockquote></details>

Could you elaborate on your comment: `it is not right.`?

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/150)*

yachang:
Review status: 0 of 1 LGTMs obtained

---

*[server/mlabns/handlers/update.py, line 268 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/150#-LFThA6T9zkCTZwRo0_k:-LFUeUPf2VwA5th6QuWz:b-19fi02) ([raw file](https://github.com/m-lab/mlab-ns/blob/ec3d8107cf8c7d6591018ed37869f33c9e19f991/server/mlabns/handlers/update.py#L268)):*

Could you elaborate on your comment: `it is not right.`?
</blockquote></details>

This IPUpdateHandler runs daily

But you said  "the delay for new data written to the datastore by IPUpdateHandler making it into memcache will be less than 60 seconds"

So the delay could be up to 24 hours, NOT less than 60 seconds.

The memcache sync code you deleted here needs to be kept, to make the datastore/memcache sync.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/150)*

nkinkade:
Review status: 0 of 1 LGTMs obtained

---

*[server/mlabns/handlers/update.py, line 268 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/150#-LFThA6T9zkCTZwRo0_k:-LFUf6dz2ncRioRO42iX:bq8k1uz) ([raw file](https://github.com/m-lab/mlab-ns/blob/ec3d8107cf8c7d6591018ed37869f33c9e19f991/server/mlabns/handlers/update.py#L268)):*

This IPUpdateHandler runs daily

But you said  "the delay for new data written to the datastore by IPUpdateHandler making it into memcache will be less than 60 seconds"

So the delay could be up to 24 hours, NOT less than 60 seconds.

The memcache sync code you deleted here needs to be kept, to make the datastore/memcache sync.
</blockquote></details>

Okay, that helps. Thank you for elaborating.

Here is why I don't believe what you are saying to be true. True, `IPUpdateHandler` only runs once a day. However, `StatusUpdateHandler` runs every 60 seconds. Starting with this PR (possibly!) `StatusUpdateHandler` reads all data from the datastore every single minute. In my mind, this means that whatever is in the datastore will be read by `StatusUpdateHandler` every minute. So, if `IPUpdateHanlder` writes a bunch of new stuff to the datastore (and nothing to memcache), `StatusUpdateHandler` will read everything that was written in something less than a minute.

If `IPUpdateHanlder` is writing to the datastore, and `StatusUpdateHandler` is reading from the datastore every minute, how is it that `StatusUpdateHandler` will have to wait up to 24 hours to get this new data? Don't forget, the whole point of this PR is that `StatusUpdateHandler` is no longer reading from memcache, but instead from the datastore.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/150)*

yachang:
Review status: 0 of 1 LGTMs obtained

---

*[server/mlabns/handlers/update.py, line 268 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/150#-LFThA6T9zkCTZwRo0_k:-LFXcChy3NwjiXMMhHq8:b-jhyyta) ([raw file](https://github.com/m-lab/mlab-ns/blob/ec3d8107cf8c7d6591018ed37869f33c9e19f991/server/mlabns/handlers/update.py#L268)):*

Okay, that helps. Thank you for elaborating.

Here is why I don't believe what you are saying to be true. True, `IPUpdateHandler` only runs once a day. However, `StatusUpdateHandler` runs every 60 seconds. Starting with this PR (possibly!) `StatusUpdateHandler` reads all data from the datastore every single minute. In my mind, this means that whatever is in the datastore will be read by `StatusUpdateHandler` every minute. So, if `IPUpdateHanlder` writes a bunch of new stuff to the datastore (and nothing to memcache), `StatusUpdateHandler` will read everything that was written in something less than a minute.

If `IPUpdateHanlder` is writing to the datastore, and `StatusUpdateHandler` is reading from the datastore every minute, how is it that `StatusUpdateHandler` will have to wait up to 24 hours to get this new data? Don't forget, the whole point of this PR is that `StatusUpdateHandler` is no longer reading from memcache, but instead from the datastore.

</blockquote></details>

Got it. Thanks!

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/150)*

yachang:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/150#-:-LFXhoyR1vz2b_fNnRF6:bnfp4nl)*

stephen-soltesz:
@nkinkade asked me to review the last edits to fix unit tests from the last comment:
https://github.com/m-lab/mlab-ns/commit/038f1d7bce61e8219ae03d56975aa970e87f462a

Those edits :lgtm:

---

Review status: 0 of 1 LGTMs obtained (and 2 stale)

---



*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/150#-:-LFZI-Mi9yg71BehTXIU:b-jef38m)*

Issue #143, opened 2018-05-30
prom queries used by mlab-ns miss lame duck values


stephen-soltesz:
During the Japan incident we discovered that HND01 could not be placed in lame duck mode because of a subtle bug in the prom query used.

The bug is the compound use of `OR` which requires that every metric have a distinct set of labels. When two or more metrics have the *same* label set then prometheus will select one (evidently the first) and discard all others.

In our query, the lame duck check comes last, and was being ignored.

As evidence:

The abbreviated query using the metrics with matching label sets (since both come from node exporter):
```
min by (experiment, machine) (
      (vdlimit_used{experiment="ndt.iupui", machine=~".*hnd.*"} /
          vdlimit_total{experiment="ndt.iupui", machine=~".*hnd.*"}) < bool 0.95 OR 
   lame_duck_experiment{experiment="ndt.iupui", machine=~".*hnd01.*"} != bool 1)
```
Returns:
```
{experiment="ndt.iupui",machine="mlab1.hnd01.measurement-lab.org"} | 1
{experiment="ndt.iupui",machine="mlab2.hnd01.measurement-lab.org"} | 1
{experiment="ndt.iupui",machine="mlab2.hnd02.measurement-lab.org"} | 1
{experiment="ndt.iupui",machine="mlab1.hnd02.measurement-lab.org"} | 1
{experiment="ndt.iupui",machine="mlab3.hnd02.measurement-lab.org"} | 1
{experiment="ndt.iupui",machine="mlab3.hnd01.measurement-lab.org"} | 1
```

Whereas when we explicitly relabel the lame duck metric to have differing label sets:
```
min by (experiment, machine) (
      (vdlimit_used{experiment="ndt.iupui", machine=~".*hnd.*"} /
          vdlimit_total{experiment="ndt.iupui", machine=~".*hnd.*"}) < bool 0.95 OR
      label_replace(lame_duck_experiment{experiment="ndt.iupui", machine=~".*hnd01.*"}, "lameduck", "foobar", "", "")   != bool 1)
```

Returns the correct values for hnd01:
```
{experiment="ndt.iupui",machine="mlab2.hnd02.measurement-lab.org"} | 1
{experiment="ndt.iupui",machine="mlab3.hnd01.measurement-lab.org"} | 0
{experiment="ndt.iupui",machine="mlab2.hnd01.measurement-lab.org"} | 0
{experiment="ndt.iupui",machine="mlab3.hnd02.measurement-lab.org"} | 1
[...]

PR #154, opened 2018-06-22
Reverts the Prometheus queries to the way they were originally created.

nkinkade:
After some thought and discussion, it was realized that the original form of the Prometheus queries is best. The current form of the queries attempts to be sure that if a metric is missing that timeseries row is still returned, with a value of zero. What this means is that when a metric is missing, the entire experiment will be flagged as down in mlab-ns. In the event there is an outage of some Prometheus service, causing a metric to disappear platform wide, this could potentially flag a tool down on the entire platform.

The old way of querying Prometheus, and the way in this PR, is to `OR` each metric we care about, and then see if any one of them equals zero. If one particular goes missing, then the other metrics should still return results, and status will be determined on the remaining metrics. If for some reason all metrics vanish, then no row will be returned at all, in which case mlab-ns will log that monitoring didn't return any results, and it will skip updating any status, leaving the current stale status.

The changes in the PR are running in mlab-nstesting.

**NOTE**: The [bug in these original queries](https://github.com/m-lab/mlab-ns/issues/143) is now handled by [a change in how we write lame duck files](https://github.com/m-lab/mlabops/pull/66).

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/154)

### DIFF
--- a/server/mlabns/util/prometheus_status.py
+++ b/server/mlabns/util/prometheus_status.py
@@ -9,6 +9,19 @@ from mlabns.util import message
 
 # This global dict maps tool_ids to the corresponding Prometheus query that
 # will return the status for that tool.
+#
+# IMPORTANT NOTE: When querying multiple metrics from the same exporter (e.g.,
+# node_exporter), be sure that the label set for each metric is unique. The `OR`
+# operator in Prometheus will exclude all metrics but one in the case where they
+# have the same label set. For example, in the queries below the vdlimit_* and
+# lame_duck_experiment metrics both come from node_exporter. By default, these
+# metrics on the same machine will have identical label sets. In this particular
+# case, to get around this, we have added unique labels to each metric via the
+# scripts that generate the node_exporter metrics files on each node. So, if,
+# for example, another blackbox_exporter metric is added (i.e., another
+# probe_success), something will need to be done to make sure that the label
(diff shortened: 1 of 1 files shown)


stephen-soltesz:
:lgtm:

---

---

*[server/mlabns/util/prometheus_status.py, line 19 at r1](https://reviewable.io/reviews/m-lab/mlab-ns/154#-LFrME68EmvuLGvVrHEb:-LFrME68EmvuLGvVrHEc:b-ltls66) ([raw file](https://github.com/m-lab/mlab-ns/blob/72fa742275229602813d88687dcad15d12058bb1/server/mlabns/util/prometheus_status.py#L19)):*
> ```Python
>             (vdlimit_used{experiment="ndt.iupui"} /
>               vdlimit_total{experiment="ndt.iupui"}) < bool 0.95 OR
>             lame_duck_experiment{experiment="ndt.iupui"} != bool 1
> ```

Please add an introductory comment advising caution when using metrics from the same exporter, and how the vdlimit and lame duck metrics are handled.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/154#-:-LFrMP5UEO25uP3UcWSe:bnfp4nl)*

PR #186, opened 2019-04-18
Change redirect to reverse proxy

stephen-soltesz:
We discovered that the 302 redirect does not work as expected for the NDT ssl client.

This change implements a basic "reverse proxy" logic where the 'redirect' logic used to be. Now, mlab-ns issues a GET request to the proxy URL and publishes the result.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/186)


stephen-soltesz:
@pboothe , Nathan requested a second reviewer for the unit tests.

Issue #189, opened 2019-05-07
On overload return 204 or 429


bassosimone:
This has been discussed by @stephen-soltesz and @bassosimone in https://github.com/m-lab/ndt-server/pull/109#pullrequestreview-233494749.

pboothe:
Code changes are in. We'd like to get positive confirmation from onebox team before we deploy.

stephen-soltesz:
The 204 change was deployed to prod yesterday. I suggest we reserve the 429 response for a future feature development.

bassosimone:
Sure, thanks!

bassosimone:
@stephen-soltesz can I close this issue now?

bassosimone:
Closing this issue, as it seems done to me. We'll reopen it if required.

PR #190, opened 2019-05-07
Respond with HTTP 204 when there is no content.

stephen-soltesz:
This change addresses part of https://github.com/m-lab/mlab-ns/issues/189

Originally, mlab-ns would return status 200 and empty result if there were no servers to report. To implement the first version of the so called "kill switch" (aka "no capacity") signal, we updated mlab-ns to return an empty json list `[]` and status 200.

In hindsight, returning a list type to some clients that request single server results is a mistake. This change instead removes the empty list response in favor of returning status 204 (No Content) when the response is empty. This allows clients to distinguish the content and for mlab-ns to treat list results and single object results more consistently.


bassosimone:
Thanks for working on this!

Issue #217, opened 2020-01-22
Build fails due to python2.7 retirement


stephen-soltesz:
The python2.7 command `pip` is no longer available for building the mlab-ns container for testing. As a result the build fails currently.

```
RUN pip install -r /test_requirements.txt
 ---> Running in b1f027805977
/bin/sh: 1: pip: not found
The command '/bin/sh -c pip install -r /test_requirements.txt' returned a non-zero code: 127
The command "docker build -t mlabns-testing ." exited with 127.
```

PR #218, opened 2020-01-23
Fix Dockerfile to remain py2.7 compatible

stephen-soltesz:
The previous base image no longer supports python2.7 natively.

This change updates the Dockerfile to remain py2.7 compatible for a little longer.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/218)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,19 @@
-FROM google/cloud-sdk
+FROM ubuntu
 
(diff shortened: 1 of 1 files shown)


pboothe:
Fixes #217 (kind of, but still leaves the 2.7 timebomb in place)

Issue #225, opened 2020-03-20
mlab-ns single-server responses should omit 1g sites


stephen-soltesz:
PR https://github.com/m-lab/mlab-ns/pull/222 introduced a static list of 1g sites to filter responses by a probability.

This is to help clients that use mlab-ns to retrieve single-server responses when the TX controller may reject measurements and the client has no additional servers to fail over to.

An even simpler strategy is to omit the 1g sites from single-server responses altogether. Conversationally this sounded like a good idea.

stephen-soltesz:
This isn't as simple as that. For Canada all sites are 1g sites, and on reflection it doesn't seem like a good idea to distort traffic in canada to travel to other regions all the time. The probability selection approach for 1g sites will reduce the traffic there and improve measurement quality when selected, but there will remain a non-zero chance that the tx controller will still reject single-server response clients (and they'll fail).

PR #226, opened 2020-03-20
Skip 1g sites for single-server responses

stephen-soltesz:
Fixes https://github.com/m-lab/mlab-ns/issues/225

For GeoResolver results requesting a single site, this change prevents ever returning 1g sites.

The rationale behind this is to prevent sending clients to tx-controller enabled 1g sites when rejected requests cannot fallback on a next server (which would be available when more than one server is returned). While it's still possible for 10g sites to reject requests the probability is much lower and the demand on such sites would be so high that most clients would be having a bad time under those conditions.

The consequence of this is client bias at 1g sites.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/226)

### DIFF
--- a/server/mlabns/util/resolver.py
+++ b/server/mlabns/util/resolver.py
@@ -131,6 +131,9 @@ class GeoResolver(ResolverBase):
                                          candidates)
 
         for candidate in filtered_candidates:
+            if max_results == 1 and candidate.site_id in site_keep_probability:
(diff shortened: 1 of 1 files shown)


stephen-soltesz:
manually deployed sandbox. And, will deploy to staging after merge.

PR #228, opened 2020-03-23
Changes to begin preparing for project-decorated/flat names.

nkinkade:
This PR touches two particular files/functions that are incompatible with project-decorated/flat hostnames. These changes were enough to get the sandbox mlab-ns records for one sample site to include a project-decorated hostnames for its corresponding `SliverTool`s.

**NOTE**: It was necessary to first manually _remove_ the sample site from the GCD, then manually run the `check_site` cron job.

**NOTE**: There are likely other places in the mlab-ns code which will break with project-decorated/flat hostnames. It is hard to fully test this in sandbox, though.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/228)


nkinkade:
Travis-CI is currently down. Merging anyway, because the Cloud Build has already run all the unit tests (all passed) and deployed this version to sandbox. The only thing Travis-CI would bring us is coverage, which will happen eventually once Travis comes back up.

PR #232, opened 2020-03-30
Changes a non-matching FQDN log message from error() to info().

nkinkade:
Not matching is not an error condition, but simply a normal process. For example, mlab1.den06 will not match in staging, and mlab4.den06 will not match in production. This tiny change helps to make the mlab-ns logs look not scary when viewing them in the GCP console.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/232)


nkinkade:
Deleted branch by accident. Reopening.

PR #236, opened 2020-05-01
Add dispatch.yaml for locate v2 paths

stephen-soltesz:
This change adds a dispatch.yaml configuration for mlab-ns.

The intended effect of the dispatch configuration is to add routing to paths on the default App Engine URL. For example, with this configuration in sandbox, the following URL routes to the locate service.

* https://mlab-sandbox.appspot.com/v2beta1/query/ndt/ndt7

I expect (but cannot confirm) that the "custom domain" on mlab-ns.appspot.com for locate.mesurementlab.net will inherit these dispatch rules. So, I expect the following to work once this config reaches the mlab-ns project.

* https://locate.measurementlab.net/v2beta1/query/ndt/ndt7

While all other current URLs continue to work as they do today.

In support of https://github.com/ooni/probe-engine/issues/562

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/mlab-ns/236)


stephen-soltesz:
https://github.com/m-lab/dev-tracker/issues/455

